PR #1, opened 2026-04-08
Add Obsidian integration: tags, aliases, and Map of Content

FrankMa1:
Add tags to concept extraction schema so the LLM returns categorical tags per concept. Enrich wiki page frontmatter with tags and deterministic aliases (slug, conjunction swap, abbreviation) for better Obsidian graph navigation. Generate a MOC.md that groups concept pages by tag with an Uncategorized fallback section.

Extract shared makeTempRoot test helper to reduce duplication.

### DIFF
--- a/src/compiler/index.ts
+++ b/src/compiler/index.ts
@@ -39,6 +39,7 @@ import {
 import { markOrphaned, orphanUnownedFrozenPages } from "./orphan.js";
 import { resolveLinks } from "./resolver.js";
 import { generateIndex } from "./indexgen.js";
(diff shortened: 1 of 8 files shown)


ethanj (approved):
Nice work. Clean separation between addObsidianMeta (field mutation) and generateMOC (pipeline), and the shared makeTempRoot fixture is a nice cleanup bonus. Alias tests cover the edge cases well.

One small observation for later: generateAbbreviation includes stop words, so 'Introduction to Neural Networks' would produce 'ITNN'. Not a blocker — we can filter common stop words if it becomes noisy.

Merging.

PR #2, opened 2026-04-08
Add multi-provider support for LLM backends

FrankMa1:
Introduce provider abstraction so users can swap LLM backends via LLMWIKI_PROVIDER env var. Supports Anthropic (default), OpenAI, and Ollama. Each provider implements a common LLMProvider interface with complete, stream, and toolCall methods.

- Add provider factory (src/utils/provider.ts) reading LLMWIKI_PROVIDER and LLMWIKI_MODEL env vars
- Add AnthropicProvider, OpenAIProvider, and OllamaProvider classes
- Refactor callClaude to delegate to the active provider while preserving retry logic and backward-compatible interface
- Replace requireApiKey with requireProvider in CLI, checking the correct API key per provider
- Add openai npm dependency
- Add fallow config to suppress interface-implementation class members
- Add tests for provider factory, OpenAI tool translation, and backward compatibility

### DIFF
--- a/.fallowrc.json
+++ b/.fallowrc.json
@@ -1,0 +1,11 @@
+{
+  "$schema": "https://fallow.tools/schema.json",
+  "overrides": [
+    {
+      "files": ["src/providers/*.ts"],
+      "rules": {
+        "unused-class-members": "off"
+      }
+    }
+  ]
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "llm-wiki-compiler",
-  "version": "0.1.0",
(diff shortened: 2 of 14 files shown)


ethanj (approved):
Good provider abstraction. The LLMProvider interface is minimal and each provider implementation is self-contained. Smart move keeping retry logic in llm.ts rather than duplicating it per provider. The requireProvider() refactor in cli.ts handles unknown providers and missing keys cleanly.

One thing to keep an eye on: the OpenAI tool schema translation only handles input_schema -> parameters at the top level. If we add nested tool schemas later we may need to recurse, but for now this covers our use cases.

PR #4, opened 2026-04-10
Add Claude settings fallback support for Anthropic provider

goforu:
## Summary
- add Anthropic credential/model/base URL fallback loading from `~/.claude/settings.json` (`env` block) while preserving explicit env precedence
- support both `ANTHROPIC_API_KEY` and `ANTHROPIC_AUTH_TOKEN`, and allow path-style custom endpoints such as `https://api.kimi.com/coding/`
- improve CLI error handling so provider validation failures are surfaced through the existing formatted command error path
- expand provider/CLI coverage with regression tests for fallback behavior and malformed Claude settings

## Verification
- npx tsc --noEmit
- npm run build
- npm test
- npx fallow --score --fail-on-issues

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ dist/
 localdocs/
 .fallow/
 .worktrees/
+.sisyphus/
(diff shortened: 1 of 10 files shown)


ethanj (approved):
Nice work. The claude-settings.ts module is well-scoped — narrow read-only integration, explicit env precedence preserved, good error handling on malformed JSON. Solid test coverage across fallback behavior, URL validation, and precedence rules.

Moving requireProvider() inside the try/catch is a good ergonomic fix — provider validation errors now route through the same formatted CLI error path as everything else.

Thanks for the README docs update.

PR #5, opened 2026-04-14
Add MiniMax as a supported LLM provider using the OpenAI-compatible endpoint

socraticblock:
Changes:
   - Add `MiniMaxProvider` class extending `OpenAIProvider` with api.minimax.io/v1 base URL
   - Add `minimax` to `SUPPORTED_PROVIDERS` and `PROVIDER_MODELS` (default: MiniMax-M2.7)
   - Add `minimax` to `PROVIDER_KEY_VARS` (uses `LLMWIKI_API_KEY` env var)

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -96,6 +96,7 @@ const PROVIDER_KEY_VARS: Record<string, string | null> = {
   anthropic: "ANTHROPIC_API_KEY",
   openai: "OPENAI_API_KEY",
   ollama: null,
+  minimax: "MINIMAX_API_KEY",
 };
 
 /** Exit with a helpful message if the selected provider's API key is missing. */
--- a/src/providers/minimax.ts
+++ b/src/providers/minimax.ts
@@ -1,0 +1,18 @@
+/**
+ * MiniMax LLM provider implementation.
+ *
+ * Extends OpenAIProvider since MiniMax exposes an OpenAI-compatible API.
+ * Overrides only the constructor to set MiniMax's base URL and API key.
+ */
+
+import { OpenAIProvider } from "./openai.js";
+
+/** MiniMax API base URL. */
+const MINIMAX_BASE_URL = "https://api.minimax.io/v1";
+
+/** MiniMax-backed LLM provider using the OpenAI-compatible endpoint. */
+export class MiniMaxProvider extends OpenAIProvider {
+  constructor(model: string, apiKey: string) {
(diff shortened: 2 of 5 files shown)


ethanj (changes requested):
Thanks for the contribution. A few changes needed before we can merge:

**1. CI is failing on fallow (codebase-health check).** The `MINIMAX_BASE_URL` export in `src/providers/minimax.ts` has no consumer — it's only used inside the constructor in the same file. Either inline it or remove the `export` keyword.

**2. Add tests.** The project has a clear pattern of testing providers — see `test/provider-factory.test.ts` and `test/provider-openai.test.ts`. Please add at least:
- Factory routing: `getProvider()` returns `MiniMaxProvider` when `LLMWIKI_PROVIDER=minimax`
- Missing key: throws a clear error when the API key env var is absent

**3. Use a provider-specific env var name.** Every other provider uses a provider-specific name (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`). `LLMWIKI_API_KEY` is too generic and suggests a global llmwiki key. Please rename to `MINIMAX_API_KEY` and update the factory + `PROVIDER_KEY_VARS` accordingly.

**4. Revert unrelated `package-lock.json` changes.** This PR removes `"peer": true` from ~8 unrelated dependencies (jsdom, esbuild, typescript, tsup, ws, etc.). This looks like a byproduct of running `npm install` with a different npm/node version. Please revert those edits — the lockfile should only reflect the `openai` peer-dep context if anything.

**5. PR title has a leading whitespace artifact.** Minor, but worth cleaning up — "   Adds MiniMax..." (3 leading spaces). Please edit the title.

Once those are addressed I'm happy to approve. Nice minimal provider implementation otherwise — extending `OpenAIProvider` is the right call given MiniMax's OpenAI-compatible API.

socraticblock:
Thanks for the thorough review @ethanj. All 5 points addressed — removed the stray export, added tests for factory routing and missing key, renamed to `MINIMAX_API_KEY`, no lockfile changes, and fixed the title whitespace. Rebased onto the latest main to resolve the merge conflicts. Should be green now.

ethanj (approved):
Thanks for the fast turnaround — all the feedback is addressed cleanly. The dedicated getMiniMaxProvider() helper matches the pattern set by getAnthropicProvider(), and the missing-key test ensures the error path stays covered. Fallow is happy, CI is green.

Merging.

PR #6, opened 2026-04-16
Add embedding-based semantic search for query routing

PipDscvr:
### Summary

- Adds embed() to the LLMProvider interface with implementations for Anthropic (Voyage voyage-3-lite), OpenAI (text-embedding-3-small), and Ollama (nomic-embed-text)
- New src/utils/embeddings.ts with cosine similarity (zero-magnitude guarded), top-K search, and .llmwiki/embeddings.json persistence
- Query integration: if embeddings exist, pre-filter the wiki index to the top 15 most similar pages before sending to the selection LLM. Builds a filtered index with stable slug identifiers, eliminating the lossy
  title-to-slug reconstruction on the fallback path. Graceful fallback when no embeddings file exists.
- Embedding store lifecycle: updated after generateIndex/generateMOC in the compile pipeline, after saveQueryPage() in the query command, and prunes orphaned entries automatically
- Anthropic path gracefully degrades if VOYAGE_API_KEY is unset — warns and falls back to full-index selection rather than failing compile

### Test plan

- 15 new embedding tests (similarity, top-K, store I/O) + 5 new provider embed tests (stubbed OpenAI client, Anthropic missing-key error)
- npm test — 193 tests pass
- fallow — clean (0 issues)

### DIFF
--- a/src/commands/query.ts
+++ b/src/commands/query.ts
@@ -19,6 +19,7 @@ import { atomicWrite, safeReadFile, slugify, buildFrontmatter, parseFrontmatter 
 import { generateIndex } from "../compiler/indexgen.js";
 import * as output from "../utils/output.js";
 import { QUERY_PAGE_LIMIT, INDEX_FILE, CONCEPTS_DIR, QUERIES_DIR } from "../utils/constants.js";
+import { findRelevantPages, updateEmbeddings } from "../utils/embeddings.js";
 
 /** Directories to search when loading selected pages, in priority order. */
 const PAGE_DIRS = [CONCEPTS_DIR, QUERIES_DIR];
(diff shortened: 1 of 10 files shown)


ethanj (approved):
Solid work. The graceful-degradation story is the standout — semantic search is wrapped in safelyUpdateEmbeddings() during compile and tryFindRelevantPages() during query, so missing VOYAGE_API_KEY or transient provider errors warn rather than break the pipeline. Cold-start logic in updateEmbeddings() is nice too: if no store exists, embed everything; otherwise only re-embed changed slugs and prune dead ones in one pass.

The slug-stable filtered index in selectRelevantPages() is exactly the right shape — passes `- **slug**: Title — Summary` lines so the tool output returns slugs directly, no slugify reconstruction needed on the semantic path. The fallback path keeps the old slugify behavior, which is the right call for backward compat.

The OllamaProvider override of embeddingModel() is a clean extension point. Merging.

PipDscvr:
> Solid work. The graceful-degradation story is the standout — semantic search is wrapped in safelyUpdateEmbeddings() during compile and tryFindRelevantPages() during query, so missing VOYAGE_API_KEY or transient provider errors warn rather than break the pipeline. Cold-start logic in updateEmbeddings() is nice too: if no store exists, embed everything; otherwise only re-embed changed slugs and prune dead ones in one pass.
> 
> The slug-stable filtered index in selectRelevantPages() is exactly the right shape — passes `- **slug**: Title — Summary` lines so the tool output returns slugs directly, no slugify reconstruction needed on the semantic path. The fallback path keeps the old slugify behavior, which is the right call for backward compat.
> 
> The OllamaProvider override of embeddingModel() is a clean extension point. Merging.

Thanks @ethanj I greatly appreciate it!

PR #12, opened 2026-04-20
Add optional split embedding endpoints for OpenAI-compatible providers

ishan5ain:
Adds support for running chat and embeddings against separate OpenAI-compatible endpoints, which is useful for local/distributed setups where a chat model and embedding model are served by different processes. The feature is optional and preserves existing behavior when no separate embedding endpoint is configured.

Key Changes

- Add optional OPENAI_EMBEDDINGS_BASE_URL support for OpenAI-compatible providers
- Add optional OLLAMA_EMBEDDINGS_HOST support for Ollama
- Add LLMWIKI_EMBEDDING_MODEL for OpenAI/Ollama embedding model overrides
- Reuse the normal chat client for embeddings when no split endpoint is configured
- Rebuild or bypass embedding stores when the active embedding model changes
- Scope LLMWIKI_EMBEDDING_MODEL to OpenAI/Ollama so Anthropic continues using its Voyage default
- Update README docs with local split-endpoint examples and optional endpoint behavior
- Add tests for split clients, provider env wiring, embedding model overrides, and model mismatch handling

Why
Some local LLM serving setups, such as llama-server, run one model per server process (or the setup requires running servers on separate host devices). That means a chat model and an embedding model often need separate OpenAI-compatible endpoints. This change makes that setup possible without requiring a new provider or changing the default behavior for existing OpenAI/Ollama users.

Tested on my local llama-server setup by following these steps:

1. Build a local fork
```bash
npm install
npm run build
```

2. Launch local llama.cpp servers

```bash
# Chat model server
llama-server \
  -hf unsloth/gemma-4-E4B-it-GGUF \
  --host 0.0.0.0 \
  --port 8080 \
  --ctx-size 32768
```
```bash
# Embedding model server
llama-server \
  -hf nomic-ai/nomic-embed-text-v1.5-GGUF \
  --host 0.0.0.0 \
  --port 8081 \
  --embedding \
  --pooling cls
```

3. Create a Temporary wiki

```bash
mkdir -p /tmp/llmwiki-split-endpoint-test
cd /tmp/llmwiki-split-endpoint-test
node path/to/llm-wiki-compiler/dist/cli.js ingest https://en.wikipedia.org/wiki/Andrej_Karpathy
```

4. Configure an OpenAI-compatible split endpoint setup (or add them to local .env in the temporary wiki): 

```bash
export LLMWIKI_PROVIDER=openai
export LLMWIKI_MODEL=unsloth/gemma-4-E4B-it-GGUF
export LLMWIKI_EMBEDDING_MODEL=nomic-ai/nomic-embed-text-v1.5-GGUF:F32
export OPENAI_API_KEY=sk-whatever
export OPENAI_BASE_URL=http://127.0.0.1:8080/v1
export OPENAI_EMBEDDINGS_BASE_URL=http://127.0.0.1:8081/v1
```

5. Run the local build directly:
```bash
node /path/to/llm-wiki-compiler/dist/cli.js compile
node /path/to/llm-wiki-compiler/dist/cli.js query "What does the source say about embedding endpoints?"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,43 @@ Example with zero exports (Claude Code already configured):
 llmwiki compile
 ```
 
+### OpenAI-Compatible Local Servers
+
+Use the OpenAI provider for local OpenAI-compatible servers such as
+`llama-server`. `OPENAI_BASE_URL` is used for chat/tool calls, and
+`OPENAI_EMBEDDINGS_BASE_URL` is optional. Set it only when embeddings are
+served from a different endpoint; when unset, embeddings use the same client
+and base URL as chat. Include `/v1` in custom URLs.
+
+Split endpoint example:
+
+```bash
+export LLMWIKI_PROVIDER=openai
+export LLMWIKI_MODEL=qwen3.6-35b
+export LLMWIKI_EMBEDDING_MODEL=text-embedding-model
+export OPENAI_API_KEY=sk-local
+export OPENAI_BASE_URL=http://host_url:port/v1
+export OPENAI_EMBEDDINGS_BASE_URL=http://host_url:port/v1
+```
+
+`OPENAI_API_KEY` is still required by the CLI and OpenAI SDK. For local
+servers that do not check authentication, any dummy value is sufficient.
+
+### Ollama
+
+Ollama uses its OpenAI-compatible endpoint. Set `OLLAMA_HOST` for chat and
+optionally set `OLLAMA_EMBEDDINGS_HOST` only when embeddings are served from a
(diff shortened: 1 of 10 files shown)


ethanj (approved):
Thanks, this is a clean PR.

Particularly liked:
- **Model-change handling** in `findRelevantPages` + `updateEmbeddings` — most contributors miss that embeddings from different models live in incompatible spaces, and the rebuild path is exactly right
- Backward-compatible options-object constructor refactor
- Tests cover the interesting cases — split clients, model mismatch rebuild, whitespace-only env vars
- README has working examples for both OpenAI and Ollama splits

Approving. Will merge shortly.

PR #15, opened 2026-04-22
fix wikilink in index

ludevica:
Fix wikilink resolution in index page by using slug-based links

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,3 @@ localdocs/
 .fallow/
 .worktrees/
 .sisyphus/
-*.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ localdocs/
 .fallow/
 .worktrees/
 .sisyphus/
+*.yaml
--- a/src/compiler/indexgen.ts
+++ b/src/compiler/indexgen.ts
@@ -100,13 +100,13 @@ function buildIndexContent(concepts: PageSummary[], queries: PageSummary[]): str
   const lines = ["# Knowledge Wiki", "", "## Concepts", ""];
 
   for (const page of concepts) {
-    lines.push(`- **[[${page.title}]]** — ${stripWikilinks(page.summary)}`);
+    lines.push(`- **[[${page.slug}|${page.title}]]** — ${stripWikilinks(page.summary)}`);
   }
 
   if (queries.length > 0) {
     lines.push("", "## Saved Queries", "");
     for (const page of queries) {
-      lines.push(`- **[[${page.title}]]** — ${stripWikilinks(page.summary)}`);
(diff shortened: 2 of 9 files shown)


ethanj:
Thanks for digging into this @ludevica, and welcome to your first contribution to the repo! You spotted something real: wiki files are stored by slug (e.g. `knowledge-compilation.md`), so `[[Title]]` doesn't reliably resolve in Obsidian unless the title also lives in `aliases`. Switching the index and MOC to `[[slug|Title]]` is the right call.

A few things to sort out before we can merge:

**The codebase-health failure was on us, not you.** The `fallow-rs/fallow@v2` action was installing the latest fallow binary, and it ticked from 2.42 to 2.45 between yesterday and today, raising complexity thresholds on two preexisting functions. We pinned the version in #16 (now merged), so once you push again the codebase-health check should pass.

**The test failures are within your scope.** The output format change broke tests that assert against the old `[[Title]]` strings. They need updating to match the new format:

- `test/indexgen.test.ts` (3 failures)
- `test/obsidian.test.ts` (4 failures)
- `test/query-save.test.ts` (2 failures)
- `test/compounding.test.ts` (1 failure)

**A couple of related things worth handling in this PR:**

1. `src/compiler/resolver.ts:116` writes `[[${page.title}]]` into page bodies. That has the same Obsidian resolution problem you fixed in the index. Worth applying the same `[[slug|Title]]` pattern there for consistency.
2. `stripWikilinks` in `src/compiler/indexgen.ts:91` uses `\[\[([^\]]+)\]\]` and replaces with `$1`. If a summary contains `[[slug|Title]]`, it would strip to `slug|Title` rather than `Title`. Probably worth handling.
3. The `*.yaml` addition in `.gitignore` looks unrelated. Mind dropping it from this PR? Easier to keep the diff focused.

Let us know if any of this is unclear or if you'd rather we take it from here. Either way, really appreciate the PR!

ludevica:
Thanks for the review — I've addressed points 1 and 3 and pushed the updates.

Regarding point 2, I'm not entirely sure what formats `summary` might contain in practice. If it can include wikilinks like [[slug|Title]], then your suggestion makes sense and I’m happy to adjust the `stripWikilinks` logic accordingly. For now, I’ve left it unchanged until we confirm the expected input format.

Let me know if you'd prefer I update it now or if we should handle it in a follow-up PR.

ethanj:
Thanks for the iterations @ludevica — both points 1 and 3 look great, and CI is fully green. The test updates cover everything cleanly.

On point 2: good question. Summaries don't actually contain wikilinks in normal operation (they're LLM-generated and the extraction prompt doesn't ask for them), so the edge case you raised is theoretical. Worst case is a cosmetic glitch in the index, not a crash.

We'll merge this as-is and handle the `stripWikilinks` regex update in a small follow-up PR ourselves. No need to ping-pong it back to you for a one-line change.

Really appreciate the contribution — looking forward to your next PR.

PR #34, opened 2026-04-27
fix(ingest): resolve youtube-transcript startup import error

### LINKED ISSUE #33: ingest fails on startup: ERR_PACKAGE_PATH_NOT_EXPORTED for "youtube-transcript/dist/youtube-transcript.esm.js"
## Summary

  `llmwiki` (any subcommand: `ingest`, `--version`, etc.) crashes on startup with
  `ERR_PACKAGE_PATH_NOT_EXPORTED` because `dist/cli.js` imports a deep subpath of
  `youtube-transcript` that no longer exists in the installed version.

  ## Environment

  - `llm-wiki-compiler`: 0.5.0 (installed via `npm install -g llm-wiki-compiler`)
  - `youtube-transcript` (resolved): 1.3.1
  - Node.js: 24.14.1
  - OS: Windows 11
  - Shell: Git Bash

  ## Steps to reproduce

  ```bash
  npm install -g llm-wiki-compiler
  llmwiki --version
  ```

  ## Actual output

  ```
  node:internal/modules/esm/resolve:314
    return new ERR_PACKAGE_PATH_NOT_EXPORTED(
           ^

  Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath
  './dist/youtube-transcript.esm.js' is not defined by "exports" in
  .../node_modules/youtube-transcript/package.json imported from
  .../node_modules/llm-wiki-compiler/dist/cli.js
      at exportsNotFound (node:internal/modules/esm/resolve:314:10)
      ...
    code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
  }
  ```

  ## Root cause

  `dist/cli.js:580` does:

  ```js
  import { YoutubeTranscript as YoutubeTranscriptUntyped }
    from "youtube-transcript/dist/youtube-transcript.esm.js";
  ```

  But `youtube-transcript@1.3.1` restructured its `dist/` to
  `dist/esm/index.js` + `dist/commonjs/index.js`, and its `package.json`
  `exports` field only exposes `.` and `./package.json` — so the deep
  subpath import fails Node's exports check, and the whole CLI fails
  to load before any user code runs.

  The new package still exports `YoutubeTranscript` from its root entry, so
  the fix is just to import from the package root.

  ## Suggested fix

  In `src/...` (or wherever the source for `dist/cli.js` line 580 lives),
  change:

  ```js
  import { YoutubeTranscript as YoutubeTranscriptUntyped }
    from "youtube-transcript/dist/youtube-transcript.esm.js";
  ```

  to:

  ```js
  import { YoutubeTranscript as YoutubeTranscriptUntyped }
    from "youtube-transcript";
  ```

  I verified this one-line patch against the published `dist/cli.js` and
  `llmwiki --version` works again. Pinning `youtube-transcript` to an older
  range in `package.json` would also work, but importing from the package
  root is forward-compatible with the current export map.

  ## Workaround for other affected users

  Until a fix is released, manually edit
  `<global-node-modules>/llm-wiki-compiler/dist/cli.js` line 580 and replace
  the subpath import with `from "youtube-transcript"`.

ishan5ain:
## Summary

Fixes a CLI startup crash caused by importing a non-exported deep path from `youtube-transcript`.

In issue #33, `llmwiki --version` and other commands failed before execution with:

`ERR_PACKAGE_PATH_NOT_EXPORTED: './dist/youtube-transcript.esm.js'`

Root cause: we imported `youtube-transcript/dist/youtube-transcript.esm.js`, but newer `youtube-transcript` versions only expose the package root via `exports`.

## What Changed

- Updated transcript ingestion import to use the package root:
  - from `youtube-transcript/dist/youtube-transcript.esm.js`
  - to `youtube-transcript`
- Removed local untyped casting/interfaces that were only needed for the deep import.
- Updated inline comments in `src/ingest/transcript.ts` to reflect `exports`-map behavior.
- Bumped dependency:
  - `youtube-transcript` from `^1.3.0` to `^1.3.1` in `package.json`
  - synced `package-lock.json` to `youtube-transcript@1.3.1`

## Why This Approach

Importing from the package root aligns with the library’s public export surface and is forward-compatible with its current module layout.

## How To Test

1. `npx tsc --noEmit`
2. `npm run build`
3. `npm test`
4. `npx fallow`

Additional smoke check:
- `node dist/cli.js --version` should print the version and not throw `ERR_PACKAGE_PATH_NOT_EXPORTED`.

## Issue

Closes #33

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "p-limit": "^6.0.0",
         "pdf-parse": "^2.4.5",
         "turndown": "^7.2.0",
-        "youtube-transcript": "^1.3.0",
+        "youtube-transcript": "^1.3.1",
         "zod": "^3.25.76"
       },
       "bin": {
@@ -4431,9 +4431,9 @@
       }
     },
     "node_modules/youtube-transcript": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/youtube-transcript/-/youtube-transcript-1.3.0.tgz",
-      "integrity": "sha512-laWv9RcKIWh6rZUH3hVnOngEvtKAhFMV5UepUO6AgevPYqe2zv8KW/uCkZJDSnPwf5/AdVu0Q66/1RDblKsp6Q==",
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/youtube-transcript/-/youtube-transcript-1.3.1.tgz",
+      "integrity": "sha512-NDCjwad113TGybbYF51y9Z4tcwzBHUZWQdF9veULNca18L+FdDbHHtTHIr69WVa3bB90l67S8kN0HtL2JO9fhg==",
       "license": "MIT",
       "engines": {
         "node": ">=18.0.0"
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "p-limit": "^6.0.0",
     "pdf-parse": "^2.4.5",
     "turndown": "^7.2.0",
-    "youtube-transcript": "^1.3.0",
+    "youtube-transcript": "^1.3.1",
     "zod": "^3.25.76"
   },
   "devDependencies": {
--- a/src/ingest/transcript.ts
+++ b/src/ingest/transcript.ts
@@ -14,25 +14,10 @@
 import { readFile } from "fs/promises";
 import path from "path";
 import { titleFromFilename, type IngestedSource } from "./shared.js";
-// The youtube-transcript@1.3.0 package ships a CJS file as its `main` entry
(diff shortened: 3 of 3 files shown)


ethanj (approved):
Thanks @ishan5ain — clean fix for a real release blocker. The deep-import workaround was specifically there because 1.3.0 had a broken main entry, and 1.3.1 closing that with a proper exports map made the workaround the bug. Switching to the root import is exactly right, and dropping the manual interface + ts-expect-error as a side benefit is nice.

CI's existing 'prints version' subprocess test catches the regression class, so the lack of a new test isn't a gap. Merging.

PR #81, opened 2026-06-05
Add Claude Agent SDK provider (local Claude Code login)

alvins82:
## What & why

Adds a new provider, **`LLMWIKI_PROVIDER=claude-agent`**, that routes LLM calls through [`@anthropic-ai/claude-agent-sdk`](https://github.com/anthropics/claude-agent-sdk-typescript)'s `query()` instead of the raw Messages API. It authenticates via the user's **local Claude Code login** (OAuth/subscription), so **no `ANTHROPIC_API_KEY` is required**.

Why do we want this - because Claude Agent SDK usage tokens comes bundled with Claude personal plans. See here - https://support.claude.com/en/articles/15036540-use-the-claude-agent-sdk-with-your-claude-plan

This fills a real gap: the existing `anthropic` provider's "zero exports" path only works if a literal key/token is present in `~/.claude/settings.json`'s `env` block — it can't use the Claude Code OAuth session. `claude-agent` works purely off the logged-in CLI.

The change is additive and slots into the existing provider pattern (`LLMProvider` interface, `getProvider()` factory, `PROVIDER_MODELS`, `provider-guard`). No pipeline, prompt, or CLI changes; all other providers are untouched.

## How it works

- **`complete` / `stream`** — single-shot generation (`MAX_TURNS` ceiling; note `maxTurns: 1` makes the SDK return an error result on *every* call, so the ceiling sits above the one turn a clean completion uses). Returns the SDK's final `result` text and appends an output-only directive to suppress agentic preamble, matching raw-API output.
- **`toolCall`** — used by compile (`CONCEPT_EXTRACTION_TOOL`), query (`PAGE_SELECTION_TOOL`), and eval (`JUDGE_TOOL`). Implemented faithfully: registers the requested tool as an in-process SDK MCP tool (`createSdkMcpServer`/`tool`) and captures the real `tool_use` input as JSON. A small JSON-Schema→Zod bridge adapts llmwiki's JSON-Schema tool definitions to the SDK's Zod-raw-shape `tool()` helper.
- **`embed`** — delegates to Voyage (Claude has no embeddings endpoint), extracted into a shared `voyage-embed` helper reused by the `anthropic` provider so there's no duplication. Compile gracefully skips embeddings when `VOYAGE_API_KEY` is unset.
- **Default model** `claude-sonnet-4-5`, overridable via `LLMWIKI_MODEL` (reuses the existing Anthropic model-resolution chain).
- **`LLMWIKI_DEBUG=1`** prints a concise per-message SDK trace (`[claude-agent] system:init`, `… assistant`, `… result:success`); `=verbose` adds the SDK's full logging. Off by default.

## Dependency bumps (forced by the SDK's peers)

- `@anthropic-ai/sdk` `^0.39` → `^0.100`
- `zod` `^3` → `^4` (zod is used in a single file with primitive types — low blast radius)

Both verified safe: the full suite passes.

## Testing

- New `test/provider-claude-agent.test.ts` — `complete`/`stream`/`toolCall`/`embed`, factory resolution (no API key required), model override, and final-result preference (mocks the SDK's `query`).
- New `test/json-schema-to-zod.test.ts` — the schema bridge: required-vs-optional, enum, nested object/array, fallback.
- `npx tsc --noEmit`, `npm run build`, `npm test` (**1215 passed**, 3 skipped), and `fallow` (0 above threshold) all green.
- **Verified end-to-end live** against a real Claude Code login (no API key): ingest → compile produces interlinked wiki pages with citations; embeddings gracefully skipped without `VOYAGE_API_KEY`.

## Reviewer notes

- `toolCall` relies on `permissionMode: 'bypassPermissions'` with `allowedTools` restricted to the single registered tool, so no file/agentic tools are reachable.
- Page generation may use up to `MAX_TURNS` (4) agent turns vs. the raw `anthropic` provider's single call; in practice a clean page is one turn.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,13 @@ instead of the raw Messages API. It authenticates with your **local Claude Code
 login** (OAuth/subscription), so **no `ANTHROPIC_API_KEY` is required** — if you
 can run `claude` in your terminal, this provider works.
 
+> **Terms of use.** This provider drives your Claude Code / Agent SDK session
+> programmatically to compile wikis. That is not automatically appropriate for
+> every account type, plan, or environment. Before using it, review Anthropic's
+> current [Claude Code](https://www.anthropic.com/legal/consumer-terms) and
+> [Agent SDK](https://docs.anthropic.com/en/api/agent-sdk/overview) terms and
+> usage policies, and make sure your intended use complies with them.
+
 ```bash
 export LLMWIKI_PROVIDER=claude-agent
 export LLMWIKI_MODEL=claude-sonnet-4-5  # optional; this is the default
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Inspired by Karpathy's [LLM Wiki](https://gist.github.com/karpathy/442a6bf555914
 - **Eval harness.** `llmwiki eval` measures health score (0–100), citation coverage and precision, optional LLM-as-judge support scoring, regression deltas, and CI-gateable thresholds.
 - **MCP server.** `llmwiki serve` exposes the full pipeline to Claude Desktop, Cursor, Claude Code, and any MCP-compatible agent — including `get_context_pack` for budgeted, citation-aware evidence packs.
 - **Bridge to runtime memory.** `llmwiki export --target json --project-id <id>` produces a typed envelope that [`@atomicmemory/llmwiki`](https://github.com/atomicstrata/atomicmemory/tree/main/packages/llmwiki) imports as one verbatim Atomic Memory record per page, preserving all advisory metadata.
-- **Provider-portable.** Anthropic, OpenAI-compatible (incl. local llama.cpp / vLLM), Ollama, GitHub Copilot.
+- **Provider-portable.** Anthropic, Claude Agent SDK (local Claude Code login, no API key), OpenAI-compatible (incl. local llama.cpp / vLLM), Ollama, GitHub Copilot.
 
 ## Who this is for
 
@@ -88,6 +88,36 @@ Example with zero exports (Claude Code already configured):
 ```bash
 llmwiki compile
 ```
+
+### Claude Agent SDK (local Claude Code login)
+
+The `claude-agent` provider routes calls through the
+[Claude Agent SDK](https://github.com/anthropics/claude-agent-sdk-typescript)
+instead of the raw Messages API. It authenticates with your **local Claude Code
+login** (OAuth/subscription), so **no `ANTHROPIC_API_KEY` is required** — if you
+can run `claude` in your terminal, this provider works.
+
+```bash
+export LLMWIKI_PROVIDER=claude-agent
+export LLMWIKI_MODEL=claude-sonnet-4-5  # optional; this is the default
+llmwiki compile
+```
+
+Notes:
+
+- Generation (`compile`) and structured extraction work off the local login with
+  no extra credentials.
+- Semantic search (`llmwiki query`) still needs embeddings, which Claude does not
+  provide. Set `VOYAGE_API_KEY` to enable them (same as the `anthropic`
+  provider); otherwise `query` falls back to lexical ranking.
+- To see what the SDK is doing, set `LLMWIKI_DEBUG=1` for a concise one-line trace
+  per SDK message (`[claude-agent] system:init`, `… assistant`, `… result:success`)
+  plus any `claude` subprocess errors. Use `LLMWIKI_DEBUG=verbose` to additionally
+  enable the SDK's full verbose logging.
+
+  ```bash
+  LLMWIKI_DEBUG=1 LLMWIKI_PROVIDER=claude-agent llmwiki compile
+  ```
 
 ### OpenAI-Compatible Local Servers
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,8 @@
       "version": "0.8.0",
       "license": "MIT",
       "dependencies": {
-        "@anthropic-ai/sdk": "^0.39.0",
+        "@anthropic-ai/claude-agent-sdk": "^0.3.163",
+        "@anthropic-ai/sdk": "^0.100.1",
         "@modelcontextprotocol/sdk": "^1.29.0",
         "@mozilla/readability": "^0.5.0",
         "chokidar": "^4.0.0",
@@ -24,7 +25,7 @@
         "sanitize-html": "^2.17.3",
         "turndown": "^7.2.0",
         "youtube-transcript": "^1.3.1",
-        "zod": "^3.25.76"
+        "zod": "^4.4.3"
       },
       "bin": {
         "llmwiki": "dist/cli.js"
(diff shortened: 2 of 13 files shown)


ethanj:
Thanks for putting this together. This is a really interesting direction, and I appreciate the amount of work here: new provider wiring, docs, tests, shared Voyage embedding helper, and the JSON-schema-to-Zod bridge. For a first contribution especially, this is a substantial PR, and there’s a lot of good plumbing here.

I do need to request changes before we can merge it. The main concern is that this provider currently fails softly in places where the rest of llmwiki tries hard to fail loudly. That can lead to empty pages, empty extractions, or incorrect eval results without a clear error.

The biggest issues I’d like to see fixed:

1. `runText()` should throw on non-success SDK result messages.

   Right now it only handles `result.subtype === "success"`. If the SDK returns auth failure, max-turns, execution failure, etc., the provider can fall through and return empty or partial text. That bypasses the normal retry/error path and can produce bad wiki output instead of stopping with a clear error.

2. `toolCall()` / `collectToolInput()` should fail if no tool call is produced.

   The current implementation asks the model to call the tool, but if it answers in prose or produces no `tool_use`, `collectToolInput()` returns fallback text or `""`. Downstream callers expect JSON. In several paths that can silently become zero extracted concepts or broken eval output. If the requested tool is not called, this provider should throw.

3. The JSON-schema-to-Zod bridge needs to preserve numeric enums correctly.

   Some tools use numeric enums, for example eval scores like `0 | 1 | 2`. The current converter turns enum values into strings via `z.enum(...)`, so numeric scores become `"0" | "1" | "2"` instead of `0 | 1 | 2`. That can corrupt eval aggregation. The converter should distinguish string enums from numeric/boolean literal enums.

4. The schema bridge should preserve field descriptions.

   The raw Anthropic path passes field-level schema descriptions through to the model. The current Zod conversion drops those descriptions, including important extraction/judging guidance. Please carry descriptions over with `.describe(...)` where available.

5. Agent SDK tool permissions need tightening.

   `allowedTools` controls approval, not tool availability. Since the SDK has built-in tools by default, this provider should explicitly pass `tools: []` for generation and structured extraction, then allow only the in-process MCP tool needed for the extraction call. This matters especially because `toolCall()` currently uses `permissionMode: "bypassPermissions"`.

6. Normal dependency install is currently broken.

   `npm ci` works from the lockfile, but `npm install --dry-run` fails with a peer dependency conflict: `openai@4.104.0` wants optional peer `zod@^3.23.8`, while this PR moves the project to `zod@^4` for the Agent SDK. We need a dependency combination that works for normal contributor installs, not only lockfile installs.

7. Add a clear terms-of-use note in the docs.

   The README currently positions this as “local Claude Code login, no API key.” That may be useful for some users, but we should not imply that it is automatically appropriate for every environment or usage pattern. Please add a short note telling users to review Anthropic’s current Agent SDK / Claude Code terms and conditions before using this provider.

The tests should also cover the failure paths above: SDK error result, no `tool_use`, prose instead of tool call, numeric enum conversion, and description preservation. The current tests cover the happy path well, but they mock around the places most likely to break in production.

Things that looked good:

- The provider registration/factory/guard wiring is clean.
- `LLMWIKI_MODEL` does reach the new provider correctly.
- The shared Voyage embedding helper is a nice cleanup and keeps Anthropic/Claude-agent behavior aligned.
- The late `VOYAGE_API_KEY` requirement is consistent with the existing Anthropic embedding path.
- No file/function size issues stood out.

I ran:

- `npm ci` ✅
- `npx tsc --noEmit` ✅
- `npm run build` ✅
- targeted provider tests ✅
- `npm install --dry-run` ❌ due to the Zod/OpenAI peer conflict

I’d be happy to keep reviewing this. The direction is promising; we just need the provider to fail loudly, preserve the tool schema contract, and avoid broad SDK tool access before it can safely land.

And Welcome to the Team!!!

alvins82:
Note - the below and the original PR was written by Claude Opus. Let me know if you mind and I will remove.

Thanks for the thorough review — really helpful, and you're right that the provider was failing soft where the rest of llmwiki fails hard. All seven points are addressed in the latest commit. Point by point:

**1. `runText()` throws on non-success results.** A shared `resultText()` helper returns text only for `subtype: "success"` and throws on every error subtype (`error_during_execution`, `error_max_turns`, `error_max_budget_usd`, `error_max_structured_output_retries`), surfacing the SDK's `errors[]`. No more falling through to empty/partial text.

**2. `toolCall()` / `collectToolInput()` throw when no tool is called.** If the stream ends without a `tool_use`, it throws and includes the model's prose (truncated) for debugging instead of returning `""`/fallback. Error results mid-stream also throw via `resultText()`.

**3. Numeric/boolean enums preserved.** The bridge now uses `z.enum(...)` only when all values are strings; numeric/boolean enums become a literal union (`z.literal`), so eval scores stay `0 | 1 | 2` rather than `"0" | "1" | "2"`. Covered by a new test.

**4. Field descriptions preserved.** Each node carries its `description` through with `.describe(...)`. Covered by a new test.

**5. Tool permissions tightened.** Both query paths now pass `tools: []` to disable all built-in SDK tools (the availability set, separate from `allowedTools`). The extraction path additionally exposes only the in-process MCP tool. Verified live that extraction still reaches the MCP tool with built-ins off.

**6. Normal install fixed.** Bumped `openai` `^4 → ^6` (peer `zod ^3.25 || ^4.0`), so `npm install` / `npm install --dry-run` resolve **without** `--legacy-peer-deps`. openai v6 required two small source tweaks: narrowing the `tool_calls` union by `type === "function"`, and passing a placeholder key when none is set (v6 throws on an empty key at construction; real validation stays in `ensureProviderAvailable`). All existing provider tests pass.

**7. Terms-of-use note.** Added to the README's `claude-agent` section, pointing users to Anthropic's current Claude Code / Agent SDK terms before using this provider.

**Tests for the failure paths:** SDK error result throws, no `tool_use` throws, prose-instead-of-tool throws, numeric-enum conversion, and description preservation.

Verification: `npx tsc --noEmit`, `npm run build`, `npm test` (1220 passing), and `fallow` (0 above threshold) all green, plus a live `compile` through the provider end to end. Rebased onto the latest branch tip. Ready for another look when you have a chance — and thanks for the warm welcome!

ethanj:
this PR required package upgrades which i welcome, but did as a separate PR.
Please update your PRs to fix conflicts.

alvins82:
Both PRs updated as requested.

ethanj (approved):
Thanks for the thorough follow-up — this is in great shape now.

Re-reviewed after the merge with main and confirmed the earlier concerns are all resolved:

- **Fails loudly now.** The provider throws on an SDK error result and when the model doesn't actually call the tool, instead of returning empty/partial text or non-JSON — so a failed run surfaces clearly rather than silently producing empty pages or zero concepts.
- **Schema bridge is faithful.** Numeric/boolean enum values keep their literal types (eval `score` stays `0|1|2`), and field descriptions carry through, so the `claude-agent` path matches the raw-API path.
- **Cleanly scoped.** With the dependency upgrades now on main, the diff is just the `claude-agent` provider, the JSON-Schema→Zod bridge, the shared `voyage-embed` helper, wiring, and tests.

Verified locally against the current deps: tsc, build, full suite, and fallow all green. I also ran a live smoke through the local Claude Code login with no API key — single-shot completion and the forced tool call both work and return parseable output.

The terms-of-use note in the README is a thoughtful addition. Approving — nice work!

PR #96, opened 2026-06-09
Add page health distribution to eval harness

dohu012:
## Health-score semantics change (please note)

This revision changes how `evaluateHealth` sources its lint rules: it now
delegates to the canonical `lint()` orchestrator via `runAllLintRules()`,
instead of maintaining a separate rule list. Diffing the two rule sets, the
delta is exactly one rule — `checkStalePages`.

**Consequence:** for wikis with freshness data, stale pages now deduct from
the corpus `health.score` (−1 each) where they previously did not. Projects
with stale pages will see slightly lower scores, and any `health_score` CI
gate could newly trip. For wikis without freshness data this is a no-op.

This removes a drift-prone duplicate (on `main`, `lint()` already included
`checkStalePages` but `evaluateHealth` did not — they had drifted). The change
is now covered by an integration test driving a wiki with actual stale pages
through `evaluateHealth` and asserting the score drops to the expected value.

### DIFF
--- a/src/eval/health.ts
+++ b/src/eval/health.ts
@@ -22,6 +22,7 @@ import {
 } from "../linter/rules.js";
 import { loadSchema } from "../schema/loader.js";
 import { countCandidates } from "../compiler/candidates.js";
+import { lint } from "../linter/index.js";
 import type { LintResult } from "../linter/types.js";
 import type { HealthResult, HealthRuleResult } from "./types.js";
 
@@ -38,7 +39,7 @@ const ERROR_RULES = new Set([
 ]);
 
 /** Compute the point deduction for a single lint result. */
-function deductionFor(result: LintResult): number {
+export function deductionFor(result: LintResult): number {
   if (result.rule === "pending-target") return 0;
   if (ERROR_RULES.has(result.rule)) return ERROR_DEDUCTION;
   if (result.rule === "contradicted-page") return CONTRADICTED_DEDUCTION;
@@ -71,23 +72,20 @@ function aggregateRules(results: LintResult[]): HealthRuleResult[] {
  * health score plus per-rule breakdown.
  * @param root - Absolute path to the project root.
  */
-export async function evaluateHealth(root: string): Promise<HealthResult> {
-  const schema = await loadSchema(root);
+/**
+ * Run all lint rules via the canonical linter orchestrator and return
+ * flat results. Delegates to lint() so the rule list stays in sync with
+ * llmwiki lint — no manual list to maintain. Used by evaluateHealth and
+ * by page-health-distribution so both see the same diagnostics.
+ */
+export async function runAllLintRules(root: string): Promise<LintResult[]> {
+  const summary = await lint(root);
+  return summary.results;
+}
 
+export async function evaluateHealth(root: string): Promise<HealthResult> {
   const [allResults, pendingReviews] = await Promise.all([
-    Promise.all([
-      checkBrokenWikilinks(root),
-      checkBrokenCitations(root),
-      checkMalformedClaimCitations(root),
-      checkOrphanedPages(root),
-      checkMissingSummaries(root),
-      checkDuplicateConcepts(root),
-      checkEmptyPages(root),
-      checkLowConfidencePages(root),
-      checkContradictedPages(root),
-      checkInferredWithoutCitations(root),
-      checkSchemaCrossLinks(root, schema),
-    ]).then((results) => results.flat()),
+    runAllLintRules(root),
     countCandidates(root).catch(() => 0),
   ]);
 
--- a/src/eval/index.ts
+++ b/src/eval/index.ts
@@ -10,12 +10,13 @@ import { evaluateHealth } from "./health.js";
 import { evaluateCitationCoverage } from "./citation-coverage.js";
 import { evaluateSourceUtilization } from "./source-utilization.js";
 import { evaluateCitationDepth } from "./citation-depth.js";
+import { evaluatePageHealthDistribution } from "./page-health-distribution.js";
 import { evaluateCitationSupport } from "./citation-support.js";
 import { collectStats, appendHistory, loadPreviousReport, loadLastFullReport } from "./stats.js";
 import { computeDelta } from "./delta.js";
 import { checkThresholds } from "./thresholds.js";
 import { ensureProviderAvailable } from "../utils/provider-guard.js";
-import type { EvalReport, HealthResult, CitationCoverageResult, SourceUtilizationResult, CitationDepthResult, CitationSupportResult, StatsResult } from "./types.js";
+import type { EvalReport, HealthResult, CitationCoverageResult, SourceUtilizationResult, CitationDepthResult, PageHealthDistributionResult, CitationSupportResult, StatsResult } from "./types.js";
 
 export const DEFAULT_SAMPLE_SIZE = 20;
 
@@ -24,13 +25,14 @@ interface EvalComponents {
   citationCoverage: CitationCoverageResult;
   sourceUtilization: SourceUtilizationResult;
   citationDepth: CitationDepthResult;
+  pageHealthDistribution?: PageHealthDistributionResult;
   stats: StatsResult;
   previousReport: EvalReport | null;
   citationSupport?: CitationSupportResult | null;
 }
 
 async function buildReport(root: string, components: EvalComponents, suite: "fast" | "full"): Promise<EvalReport> {
-  const { health, citationCoverage, sourceUtilization, citationDepth, stats, previousReport, citationSupport } = components;
+  const { health, citationCoverage, sourceUtilization, citationDepth, pageHealthDistribution, stats, previousReport, citationSupport } = components;
   const partial = {
     suite,
     timestamp: new Date().toISOString(),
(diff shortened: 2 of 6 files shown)


ethanj (changes requested):
Thanks for this. The core evaluator is close: the per-page scoring idea fits the issue #78 direction, the tier-boundary tests are useful, and reusing the existing health deductions is the right direction.

I’d like to get a few issues fixed before merging:

1. `src/eval/report.ts` is based on a stale copy and reverts recently merged formatter fixes. Current `main` has an ANSI-aware/truncating `line()` helper and renders source-inventory warnings via `warningRows()`. This PR replaces that with the older `padEnd` implementation, removes the terminal warning output, and deletes the regression tests that pinned both behaviors. Please rebase this file onto current `main`, preserve the existing `line()` / `warningRows()` behavior and tests, then add the Page Health section on top.

2. `pageHealthDistribution` is required on `EvalReport`, which breaks rendering old eval history. `llmwiki eval report` re-renders stored reports, and reports written before this PR won’t have the new field, so `formatPageHealthDistribution()` can throw when it reads `perPage`. Please make the field optional/backward-compatible, like `citationSupport`, and skip the section when it’s absent.

3. The new metric does not include the current freshness lint rule. `llmwiki lint` now runs `checkStalePages`, but `runAllLintRules()` in `src/eval/health.ts` maintains a separate lint list and omits it. That means a stale page can show as healthy in the new Page Health distribution. Please either delegate to the canonical lint orchestrator or update the shared lint path so eval health and page health include the same rules as `llmwiki lint`.

A couple of should-fix items while you’re in there:

- The Page Health terminal rows are too wide for the 49-char box. Even with truncation restored, the distribution row and worst-page rows will lose useful information. A more compact layout, for example slug + score on one line and issues on a second dimmed line, would be easier to read.
- The eval pipeline now runs the lint rules twice, once for corpus health and once for page health. If it stays this way for PR size, that’s survivable, but since this is all internal plumbing it would be cleaner to run the lint pass once in `runEval()` and feed the same results to both consumers.

Nits: please use named constants for the 90/70/50 tier cutoffs, and consider avoiding the `needs_work` vs `needsWork` mismatch in the JSON shape if it’s still easy to adjust.

The evaluator core looks salvageable; the main thing is to preserve current `main` behavior and make the new report field backward-compatible.

dohu012:
Thanks for the detailed review, Ethan — all six points addressed. Rebased the whole branch onto current main to fix the stale-base problem.
1. report.ts — Rebased onto current main. The ANSI-aware line() / warningRows() behavior and their regression tests are preserved untouched; I only added formatPageHealthDistribution plus one call site in the terminal report. Confirmed via diff that nothing upstream was removed.
2. Backward compatibility — pageHealthDistribution is now optional on EvalReport (same pattern as citationSupport), and the formatter skips the section when the field is absent, so re-rendering old eval history no longer throws.
3. Unified lint source — runAllLintRules now delegates to the canonical lint() orchestrator instead of maintaining its own list, so eval health and page health include the same rules as llmwiki lint, checkStalePages included. This also means lint runs once and feeds both consumers (addresses the double-run point below).
One semantics note worth flagging: because health now goes through lint(), it includes checkStalePages. For wikis with no freshness data this is a no-op — the overall health score is unchanged. For wikis with freshness data, stale pages now contribute (−1 each via DEFAULT_DEDUCTION), so the corpus health score can be slightly lower than before. This is a direct consequence of unifying the lint source. If you'd rather keep freshness out of the corpus score, I can have deductionFor return 0 for stale-page while still surfacing it in per-page health — your call.
(Transparency: I couldn't exercise the stale path end-to-end on my test wiki — it has no state.json, so the freshness snapshot resolves to missing and checkStalePages is skipped. But page-health consumes the same lint() output as llmwiki lint, so stale pages flow through the identical path, and there's now a unit test pinning the stale-page deduction.)
4. Terminal layout — Compacted: slug + score on one line, issues on a dimmed second line, fits the 49-char box.
5. Double lint run — Resolved by #3; lint() runs once.
6. Nits — Tier cutoffs are now named constants (TIER_HEALTHY / TIER_ADEQUATE / TIER_NEEDS_WORK); JSON shape uses needs_work consistently.
Tests: 87 pass locally including a new stale-page deduction test, tsc clean. Ready for another look.

ethanj (changes requested):
Thanks for the rework — this revision is in much better shape. Both blocking issues from the earlier pass are resolved: the report renderer is back to ANSI-aware width measurement with truncation (frame stays intact), and `pageHealthDistribution` is optional, omitted when absent, produced defensively, and guarded in the formatter. Tier cutoffs are named constants and the `needs_work` key is consistent. Local gates are green: tsc, build, full suite (1763 passed), fallow (maintainability 91.2, dupes clean), and CI is passing.

A few things to address before merge — two correctness issues and a behavior change worth surfacing.

## Health-score semantics changed (please call out + cover with a test)

The largest change in this revision isn't the new metric — it's that `evaluateHealth` no longer runs its own rule list and now delegates to the canonical `lint()` orchestrator via `runAllLintRules()`. Diffing the two rule sets, the delta is exactly one rule: `checkStalePages`. So stale pages now deduct from the corpus `health.score` where they previously did not.

This is defensible — it removes a duplicate, drift-prone rule list (on `main`, `lint()` already included `checkStalePages` and `evaluateHealth` didn't, so they had drifted). But it silently changes the project's headline metric inside a PR scoped to "page health distribution," and the original description said overall health semantics were unchanged. Two asks:

- Call it out explicitly in the PR description (projects with stale pages will see lower scores; `health_score` CI gates could newly trip).
- Add an integration test that drives a wiki with actual stale pages through `evaluateHealth` and asserts the score drops. Right now nothing pins this — the existing `eval-health.test.ts` fixtures don't trigger staleness, which is why the rule-set swap landed with all `score` assertions still green. The new `deductionFor("stale-page")` unit test only covers the deduction in isolation, not the integration.

## Old eval history now renders false data (medium)

`src/eval/report.ts:180` treats a missing `pageHealthDistribution` the same as an evaluated empty wiki and prints `Page Health: (no pages)`. For a report recorded before this PR, that field being absent doesn't mean "no pages" — it means the metric didn't exist yet. `llmwiki eval report` re-renders stored history, so this will mislabel real wikis. Return `[]` when `d` is absent, and only print `(no pages)` when `d` exists with `perPage.length === 0`.

## runEval silently drops page-health failures (medium)

`src/eval/index.ts:60` wraps `evaluatePageHealthDistribution(root).catch(() => undefined)`, converting any error into a missing metric. That conflicts with the repo's fail-loud posture (no silent catches, no degraded fallbacks) and could ship an eval report quietly missing a dimension. Old-history compatibility is already handled in the type and formatter, so a fresh eval run should either produce the metric or fail — drop the catch.

## Distribution + worst-page rows don't fit the box (low)

`src/eval/report.ts:187` puts all four tier counts on one line, which exceeds the 49-col frame; the truncation then hides the `broken` count — the most important tier. The worst-page rows have the same problem: rendering a realistic report, the score and issues get truncated away (`concepts/distributed-consensus-algorithms …`), so the actionable payload is invisible in a real terminal. Split the distribution across two shorter lines (or a compact label format), and give worst pages a two-line layout (slug + score, then issues dimmed) or trim to basenames.

## Worst-page ordering isn't stable (low)

`src/eval/page-health-distribution.ts:80` sorts only by `score`, so equal-score pages fall back to Map insertion order (effectively filesystem/lint order). Tie-break by `slug` for deterministic JSON and history output.

## Minor

`runEval` calls `evaluateHealth` and `evaluatePageHealthDistribution` concurrently, and each invokes `runAllLintRules()` → `lint()`, so all rules plus `loadSchema` and the freshness snapshot run twice per eval. You flagged this as an acceptable file-I/O trade-off, which is fair; just noting a single-pass refactor is straightforward since `evaluateHealth`'s signature is internal.

Happy to re-review quickly once these are in — the core evaluator is solid and the `lint()` delegation is a nice simplification.

dohu012:
Thanks for the thorough re-review — all addressed.

**Health-score semantics:** Called out explicitly in the PR description now
(stale pages lower the corpus score; `health_score` gates may newly trip).
Added an integration test that drives a wiki with real stale pages (via
`writeSourceState` with a mismatched hash) through `evaluateHealth` and
asserts the score drops to the exact expected value (99 for one stale page),
plus a page-health assertion that the stale page is no longer scored as a
healthy 100 — pinning the original bug as fixed.

**Old eval history:** Fixed — `report.ts` now returns `[]` when the field is
absent (metric didn't exist yet), and only prints `(no pages)` when the field
exists with `perPage.length === 0`.

**Silent catch:** Dropped. A fresh eval run now either produces the metric or
fails; old-history compatibility stays handled by the optional type + formatter
guard.

**Terminal layout:** Distribution split across two lines and worst pages given
a two-line layout (slug + score, issues dimmed) so the broken count and scores
no longer get truncated out of the 49-col frame.

**Worst-page ordering:** Now tie-breaks by slug (`localeCompare`) for
deterministic output.

**Double lint run:** Left as-is per your note — keeping it as the acceptable
file-I/O trade-off for this PR's scope.

Local: tsc clean, full suite green (incl. the new stale integration tests).
Ready when you are.

ethanj (approved):
Approving — this is in great shape, and every point from the last round is resolved.

Nice work, and thanks for the thorough back-and-forth on the health-score semantics — disclosing and pinning that was the right call. 🚀

dohu012:
Thanks Ethanj — appreciate the thorough reviews and guidance through both PRs.

PR #98, opened 2026-06-14
feat(embeddings): Add batch API, binary vector store, and modularize pipeline

bdogabriel:
Split the embedding pipeline into focused modules and added two opt-in features behind env vars:
- `LLMWIKI_BATCH_EMBEDDINGS=true` — groups page and chunk texts into single API calls instead of embedding one-by-one
- `LLMWIKI_BINARY_EMBEDDINGS=true` — packs vectors into a compact Float32 .bin file instead of JSON to avoid Node's string-length ceiling

Both default to off, so the existing sequential pipeline and JSON store are unchanged.

---

I loved the project and I wanted to test it in my company's documentation repo. But it was over 900 markdown files, generating over 33000 chunks to compile. The compilation was taking ages, so I implemented batching and saving the embeddings as binary to avoid json string size limitations. These features are controlled with env vars and the code changes don't affect the current default behavior of the CLI.

It worked like a charm for my use case: compilation orders of magnitude faster and query, view and context commands working as expected.

I didn't do extensive manual testing, didn't test the MCP too, because I was mostly experimenting and messing around. But I figured I should open the PR because maybe this is something you or some other people with similar use cases might appreciate.

Issue: #99

### DIFF
--- a/src/commands/query.ts
+++ b/src/commands/query.ts
@@ -30,9 +30,9 @@ import {
 import {
   findRelevantPages,
   findRelevantChunks,
-  updateEmbeddings,
   type ChunkEmbeddingEntry,
 } from "../utils/embeddings.js";
+import { updateEmbeddingsDispatched as updateEmbeddings } from "../utils/embeddings-dispatch.js";
 import { rerankWithBm25 } from "../utils/retrieval.js";
 import { appendLog, formatWikilinkList } from "../utils/activity-log.js";
 import type { ChunkCitation, QueryResult, RetrievalDebug } from "../utils/types.js";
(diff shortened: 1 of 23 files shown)


ethanj:
This is exactly the kind of scale issue we want to solve. The prototype is very helpful. I think we should split it into batching first and binary storage second.
Would you consider working on this more, or are you just handing it off?

bdogabriel:
I could keep working on it, I just can't guarantee it'll be done soon, haha. I've been short on time lately, so if anyone wants to pick it up too, I have no problem with that!

ethanj:
i think this is a high priority, so I am going to close and use your PR as a reference implementation.

PR #121, opened 2026-06-20
feat(profile): Phase 3 PR2 — default raw-page mutation support (Unicode-safe)

ethanj:
Stacked on PR #120 (do not merge yet). Lets the write planner handle DEFAULT pages, whose slugs come from `slugify` (Unicode) and per the Phase-1 invariant never become typed entity ids.

- New `isSafeFilenameComponent` identity floor for default pages: allows Unicode letters/digits/hyphens, rejects path separators, NUL, dot-only, and leading-dot (hidden) names — distinct from the ASCII slug-safe grammar used to mint entity ids.
- New `RawPageRef` target + `planDefaultPageMutation`: plans a default page write from a raw `<dir>/<slug>.md` path with no entity id, running the same mandatory floor (collision/resource/frontmatter) and create/update intent as the typed path. The shared floor→compose→build core is factored so the typed and default paths don't duplicate it.
- The executor derives the path for either target shape and re-asserts the filename floor for raw pages (defense-in-depth).

This is the prerequisite for routing the default review-approve write through the planner without refusing non-ASCII pages. Still unwired; default profile byte-identical, full suite green.

### DIFF
--- a/src/profile/identity.ts
+++ b/src/profile/identity.ts
@@ -30,6 +30,24 @@ export function isSlugSafe(s: string): s is SlugSafe {
 }
 
 /**
+ * The DEFAULT-page identity floor: is `s` a single safe path component for a
+ * default wiki page filename?
+ *
+ * Distinct from the slug-safe EntityId grammar ({@link isSlugSafe}): default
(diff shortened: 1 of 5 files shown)


ethanj:

PR #122, opened 2026-06-20
feat(profile): Phase 3 PR3 (keystone) — route review-approve through the write planner

ethanj:
Stacked on PR #121 (do not merge yet). The KEYSTONE: gives the write planner its first real production consumer and proves the parity-preserving executor.

- `review approve` now plans its page write through `planDefaultPageMutation` and applies it through the executor (under the review command's existing lock, via a new lock-free `applyApprovedMutationsLocked` so there's no nested-lock deadlock). Same bytes, same path, same stdout/exit — only the write mechanism changes.
- The mandatory floor (path-confinement, collision, resource-limit, frontmatter) now governs an approval; a body that fails it makes approval fail loudly with the candidate retained, rather than writing unchecked.
- `replayJournal` now runs on the approve path, so a crash mid-write is recovered under the review lock.

Every existing review-approve test passes unchanged (the parity proof); default profile byte-identical; full suite green. A follow-on can extend the same rerouting to the compile batch write.

### DIFF
--- a/src/commands/review-approve.ts
+++ b/src/commands/review-approve.ts
@@ -14,10 +14,9 @@
  */
 
 import path from "path";
-import {
-  atomicWrite,
-  validateWikiPage,
-} from "../utils/markdown.js";
+import { validateWikiPage } from "../utils/markdown.js";
+import { planDefaultPageMutation } from "../trust/planner.js";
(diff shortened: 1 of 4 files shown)


ethanj:

PR #123, opened 2026-06-20
fix(state): Phase 3 PR4 — tolerant v2 migration (non-ASCII slugs) + mirror sync at writeState

ethanj:
Stacked on PR #122 (do not merge yet). Two state-v2 correctness fixes from the audit (latent until typed ownership is wired, but they de-risk it).

- **Tolerant migration:** v1→v2 migration previously threw on any non-ASCII slug (the typed-id grammar is ASCII, but `slugify` keeps Unicode), aborting the whole upgrade for non-English wikis. It now types only slug-safe slugs and keeps the rest verbatim in the `concepts`/`frozenSlugs` mirror (no typed id). Decision recorded: widening the entity-id grammar to Unicode is a deferred, broader identity change.
- **Mirror sync:** the typed `entities`/`frozenEntities` mirror is now re-derived from the v1 fields at the `writeState` choke point, so it can't desync when a v2 state is written by a writer that only knows v1 fields (which previously left stale/wrong `frozenEntities`).

Strict no-op for v1 (default), so default profile byte-identical; full suite green.

### DIFF
--- a/src/state/migrate.ts
+++ b/src/state/migrate.ts
@@ -13,11 +13,14 @@
  *    migration never double-types a slug (no `concepts/concepts/rag`);
  *  - deterministic: minted id arrays are sorted lexicographically so repeated
  *    migrations of equal inputs serialise byte-for-byte identically;
- *  - fail-closed: a bare slug that violates the slug-safe grammar throws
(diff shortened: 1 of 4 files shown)


ethanj:

PR #125, opened 2026-06-20
feat(cli): Phase 3 PR5 — state reset recovery command + user docs

ethanj:
Stacked on PR #123 (do not merge yet). Closes the product gap the audit flagged: a state file written by a newer llmwiki version fails closed across every command, but a user on an older pin had no in-app way forward.

- **`llmwiki state reset`** — backs `.llmwiki/state.json` up to `state.json.bak` and removes it so the next compile starts fresh. Refuses by default and prints exactly what it will do; `--yes` applies. The reset operates on the raw bytes (never parses the state), so it works even on a too-new or corrupt file — which is the whole point.
- **First user docs + CHANGELOG** for the fail-closed behavior and the recovery procedure (public-facing; no internal names). The documented error string matches the real one verbatim.

Subprocess tests cover all four contract paths (dry refusal, confirmed reset, too-new recovery, no-file no-op). Default profile byte-identical; full suite green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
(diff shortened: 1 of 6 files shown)


ethanj:

PR #127, opened 2026-06-21
fix(trust,cli,sdk): Phase 3 stack hardening — close review-queue + trust-boundary holes

ethanj:
Top of the Phase 3 stack (do not merge yet). Closes five findings from an independent review of the full stack before it's merge-ready.

- **Typed candidates approve to the right place.** The canonical `review approve <id>` previously ignored a candidate's `targetEntityType`, so a staged `papers` candidate landed in `wiki/concepts/`. Approve now loads and validates the active profile and routes a typed candidate through the typed planner to `wiki/<entityType>/<slug>.md` (or refuses if the project has no such entity type) — and the default concepts/queries path stays byte-identical.
- **Candidate ids can't escape the candidate store.** Candidate ids were slug-derived path segments, so an unsafe slug could traverse out of `.llmwiki/candidates/`. Ids are now validated as single safe path components and confined before every read/write/delete/archive; an unsafe slug is refused at creation, and a blocked stage no longer persists a candidate at all.
- **Promotion is race- and profile-safe.** Staged promotion now re-reads the candidate, re-validates the active profile, re-plans, applies, and deletes — all under one project lock — so a concurrent reject or a profile change after staging can't promote into a stale target. The SDK helper and `review approve` share one under-lock routine.
- **`state reset` is locked and confined.** It now acquires the project lock and validates `.llmwiki`'s realpath stays in the project (failing closed on a symlink escape) before touching anything — the same containing-directory trust boundary recently fixed for the journal. Confinement is checked before the lock so a symlinked `.llmwiki` can't even get a lock file written through it.
- **The staging loop is reachable.** `createWiki()` gains experimental `stageEntityPage`/`promoteStagedPage` methods that load the active profile internally, plus the public types, so non-default staging is usable by package consumers.

Default profile byte-identical; existing review-approve tests pass unchanged; full suite green.

### DIFF
--- a/src/commands/review-approve.ts
+++ b/src/commands/review-approve.ts
@@ -22,6 +22,7 @@ import {
   CandidateProfileError,
   CandidatePromotionBlockedError,
 } from "../trust/promote.js";
+import { EntityFieldContractError } from "../profile/field-contract.js";
 import { deleteCandidate } from "../compiler/candidates.js";
 import { generateIndex } from "../compiler/indexgen.js";
 import { generateMOC } from "../compiler/obsidian.js";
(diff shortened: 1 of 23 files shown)


ethanj:

PR #128, opened 2026-06-21
fix(trust,cli): Phase 3 audit bug fixes — dedup, flood bound, review-queue visibility, query-save, races

ethanj:
Stacked on PR #127 (do not merge yet). Closes the must-fix bug findings from a four-lens principal audit of the Phase 3 stack. Default profile byte-identical; existing tests pass unchanged.

- **Data loss fixed:** candidate dedup keyed on slug ONLY collapsed two different-entity-type candidates sharing a slug into one (silently dropping a staged page). Dedup now keys on the full target identity `(entityType/directory, slug)`; default concepts behavior unchanged.
- **Flood DoS fixed:** the per-session staged-write cap trusted a caller-supplied count the SDK defaulted to 0, so it never bit. It now reads the on-disk candidate count, so the cap is real.
- **Parse-before-cap fixed:** the field-contract YAML parse ran before the resource-limit check on the staging/promote path; a raw length guard now rejects oversized bodies before any parsing.
- **Human-gate visibility:** `review list`/`review show` now surface a candidate's typed target (`wiki/<entityType>/<slug>.md`), so a reviewer can see where a staged typed page will land instead of it looking like a concept.
- **query --save** is now refused in profile-enabled projects (its saved-page write isn't trust-routed yet), per the trust contract — the answer still prints; default projects are unchanged.
- **Confine→write race closed:** the shared atomic-write primitive now refuses a symlinked target directory, closing a local-tamper window between path confinement and the write across every confined caller.
- **Graceful failures + honest docs:** `state reset` gives an actionable message instead of a raw EISDIR when the backup path is a directory; approve/executor docstrings now state only the page write is journalled (the post-write tail is best-effort/idempotent); freshness documents that typed pages are intentionally unverified this phase.

Full suite green (2020).

### DIFF
--- a/src/commands/query-save.ts
+++ b/src/commands/query-save.ts
@@ -1,0 +1,114 @@
+/**
+ * @file src/commands/query-save.ts
+ * @description The `query --save` write path — persists a generated answer as a
+ * `wiki/queries/<slug>.md` page and refreshes the index/embeddings so the answer
+ * is immediately retrievable.
+ *
+ * Split out of `query.ts` to keep that command file within the project size
+ * budget; the answer-generation pipeline stays in `query.ts` and calls
+ * {@link maybeSaveQueryPage} once the answer is produced.
+ */
+
+import path from "path";
+import { atomicWrite, slugify, buildFrontmatter } from "../utils/markdown.js";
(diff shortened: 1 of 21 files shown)


ethanj:

PR #130, opened 2026-06-21
refactor(lint,candidates): Phase 4 groundwork — split rules.ts and candidates.ts under the size limit

ethanj:
First of the stacked Phase 4 PRs (do not merge yet). Pure mechanical groundwork: both `src/linter/rules.ts` (~643) and `src/compiler/candidates.ts` (~442) were over the 400-line code limit and are exactly where Phase 4's relation/lifecycle lint and candidate logic will land. Split now, before adding logic.

- `candidates.ts` → `candidate-read.ts` (read/list/sanitize/validate) + `candidate-paths.ts` (shared path resolvers), with `candidates.ts` keeping the write/dedup side and re-exporting the rest so all import sites are unchanged.
- `rules.ts` → `rules-citations.ts` + `rules-crosslinks.ts` + `rules-shared.ts`, with `rules.ts` keeping the wikilink/quality rules and re-exporting the families.

Leaf shared-modules break the re-export/import cycles. Zero behavior change: lint codes/messages and candidate output identical, default goldens byte-identical, full suite unchanged (2145).

### DIFF
--- a/src/compiler/candidate-paths.ts
+++ b/src/compiler/candidate-paths.ts
@@ -1,0 +1,65 @@
+/**
+ * Shared candidate-store path resolution for the review pipeline.
+ *
+ * Both the WRITE side ({@link file://./candidates.ts}) and the READ side
+ * ({@link file://./candidate-read.ts}) of the candidate store need to turn a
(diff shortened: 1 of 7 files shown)


ethanj:

PR #131, opened 2026-06-21
feat(profile): Phase 4 PR2 — runtime lifecycle transition enforcement + invalid-state lint

ethanj:
Stacked on PR #130 (do not merge yet). The lifecycle FSM *definition* was already validated at profile-load; this adds the *runtime* enforcement the spec requires ("direct writes / review approval / MCP / SDK cannot perform illegal transitions; required evidence enforced by core").

- New `validateLifecycleTransition` (sibling of the field-contract validator): rejects an undeclared state, an illegal transition (next not in the legal out-edges of prev; a terminal prev has none), and a transition to a state missing its declared `transitionRequirements` evidence. New-page creates are exempt from the edge check but still gated on state-validity + evidence.
- Wired as a mandatory, fail-closed gate into every typed write — staging, promotion, and `review approve` — so an illegal transition is refused (candidate retained), sharing one implementation across CLI/SDK.
- An `invalid-lifecycle-state` lint flags a typed page whose lifecycle field isn't a declared state.

Default/concepts declare no lifecycle, so the gate is a no-op there — default profile byte-identical; full suite green (2155).

### DIFF
--- a/src/commands/review-approve.ts
+++ b/src/commands/review-approve.ts
@@ -33,6 +33,7 @@ import {
   CandidatePromotionBlockedError,
 } from "../trust/promote.js";
 import { EntityFieldContractError } from "../profile/field-contract.js";
+import { LifecycleTransitionError } from "../profile/lifecycle.js";
 import { deleteCandidate } from "../compiler/candidates.js";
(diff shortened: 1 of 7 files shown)


ethanj:

PR #132, opened 2026-06-21
feat(profile): Phase 4 PR3 — typed relation type definitions + fail-closed load validation

ethanj:
Stacked on PR #131 (do not merge yet). Lets a non-default profile DECLARE typed relation types (the `relations` block) — schema + validation only; the relation store and writes come in PR4/PR5.

- New `RelationTypeDef` (`from`/`to` endpoint entity-type lists, `direction` directed|symmetric, typed `attributes`, `requiredAttributes`) + an optional `relations` block on the profile, in the type, the JSON schema, and the digest.
- Fail-closed load validation: a relation whose `from`/`to` references an undeclared entity type, a `requiredAttributes` entry not in `attributes`, a slug-unsafe relation key, or an invalid `direction` all reject the profile. `profile show` surfaces declared relations.

The default profile declares no relations, so the block is omitted everywhere — default byte-identical; full suite green (2167).

### DIFF
--- a/src/commands/profile.ts
+++ b/src/commands/profile.ts
@@ -41,6 +41,21 @@ export async function profileShow(): Promise<void> {
   console.log(`profileId:  ${loaded.profile.profileId}`);
   console.log(`digest:     ${loaded.digest}`);
   console.log(`loadedFrom: ${loadedFrom}`);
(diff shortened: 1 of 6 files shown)


ethanj:

PR #133, opened 2026-06-21
feat(relations): Phase 4 PR4 — append-only JSONL relation store + RelationRef

ethanj:
Stacked on PR #132 (do not merge yet). The durable typed-relation store; not yet wired through the planner (PR5).

- `RelationRef` (`rel_<ULID>` id allocated once, never recomputed; `contentHash` = canonical RFC8785 digest over {type,from,to,attributes,evidence}; symmetric edges canonicalized so a↔b hashes identically).
- `wiki/graph/relations.jsonl` with the JSONL durability contract: header `schemaVersion` (reader fails closed when newer), per-record checksum (interior corruption fails closed), torn trailing line tolerated + reported (never silently dropped), appends under the project lock. The graph dir is realpath-confined (fails closed on a symlink escape, mirroring the journal).
- Endpoint validation against the profile relation def (from/to entity types + required attributes) before any write. Dependency-free ULID via node:crypto.

A default project has no `wiki/graph/` → empty relations, parity intact; full suite green (2182).

### DIFF
--- a/src/relations/digest.ts
+++ b/src/relations/digest.ts
@@ -1,0 +1,68 @@
+/**
+ * @file src/relations/digest.ts
+ * @description Canonical content hashing for relations. Mirrors
+ * `src/profile/digest.ts`: a SHA-256 over the RFC 8785 (JCS) canonicalization
+ * of the content fields, so equal content yields an equal hash regardless of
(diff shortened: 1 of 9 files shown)


ethanj:

PR #134, opened 2026-06-21
feat(trust): Phase 4 PR5 — trust-gated relation writes + lifecycle transition API

ethanj:
Stacked on PR #133 (do not merge yet). Gives relations and lifecycle transitions a trust-gated write API (Invariant 4: writes route through one planner decision), shared by SDK/CLI.

- `createRelation` plans through `planRelationMutation` (endpoint + required-attribute checks → a composed deny on violation) then appends via the relation store's own append-only durability under the lock — never the page journal. A denied write throws and writes nothing.
- `transitionLifecycle` is a thin, correctly-gated read-modify-write: it sets the page's lifecycle field + evidence and writes through the existing typed page path, which re-runs the PR2 transition gate, so an illegal transition or missing evidence is refused and the page is unchanged.
- Both exposed as `@experimental` `createWiki()` methods that load the profile internally; the page executor/journal is untouched (the relation/lifecycle-transition MutationKinds never reach the page `applyBatch`).

Default projects (no relations/lifecycle) can't call these meaningfully and are unaffected — default byte-identical; full suite green (2192).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -63,14 +63,29 @@ export { LockUnavailableError, QueueFullError } from "./import/run-errors.js";
 // in interlinking, semantic search/embeddings, the MOC, or the viewer (planned).
 export type { SdkStageEntityPageInput } from "./trust/staging.js";
 export { StagingRequiresProfileError } from "./trust/staging.js";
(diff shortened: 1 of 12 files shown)


ethanj:

PR #135, opened 2026-06-21
feat(relations): Phase 4 PR6 — dangling/corrupt relation lint + relation counts/views in status & export

ethanj:
Top of the Phase 4 stack (do not merge yet). Makes the relation store visible.

- **Lint:** `checkRelationStore` reads the store fail-closed — a corrupt/too-new store surfaces a lint finding (never crashes lint), a torn trailing line surfaces a warning, and a relation whose from/to endpoint has no page is flagged `dangling-relation`.
- **Status/viewer:** the additive profile block gains optional `relationCounts`/`relationTotal` (per relation type), populated only when a non-default profile has a non-empty store; a fail-closed read surfaces a problem instead of crashing status.
- **JSON export:** the profile block gains an optional, path-safe `RelationView[]` DTO (id/type/from/to/attributes/contentHash — no paths).

Every new field is optional and omitted-for-default, so a default project and a relation-less non-default project emit no relation fields — status/export/lint byte-identical; full suite green (2203).

Deferred (noted): viewer graph filters and context-pack relation expansion.

### DIFF
--- a/src/export/json-export.ts
+++ b/src/export/json-export.ts
@@ -31,6 +31,33 @@ import type { ExportPage } from "./types.js";
 import type { EntityPageView, EntityProblemView } from "../profile/types.js";
 
 /**
+ * The PUBLIC, path-safe DTO for one typed relation in the JSON export.
+ *
+ * Mirrors the {@link EntityPageView} omitted-for-default discipline: it NEVER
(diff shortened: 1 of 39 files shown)


ethanj:

PR #136, opened 2026-06-21
fix(security): harden atomicWrite + all temp-writers against the systemic leaf-symlink write-escape

ethanj:
Stacked on PR #135 (do not merge yet). The symlink-write-escape class recurred FIVE times in the relation store (dir → compaction temp → store-file leaf), each fixed there. This closes the SAME class systemically across every other store, at the shared primitive.

The vector: every bespoke temp+rename writer did `writeFile(<path>.tmp, content)` then `rename` — and `writeFile` follows a symlinked `<path>.tmp`, so a pre-planted symlinked temp leaf redirects the write outside the project root (identical to the relation-compaction-temp bug).

- **`atomicWrite`** (the shared primitive, used by 16 files: page writes, index, MOC, candidates, …) now writes through a RANDOM-named temp opened with `open("wx")` (O_EXCL — refuses any pre-existing entry including a symlink → fail closed), then renames. The parent-dir lstat guard stays. Fixing the primitive once hardens every caller.
- **`writeState`**, **`writeRuleState`**, and the journal **`persist`** — which each had their own unguarded temp+rename — now route through the hardened `atomicWrite` (DRY).

Default page-write path byte-identical (a random O_EXCL temp produces identical destination bytes for a normal write — frozen goldens unchanged); full suite green (2267). Reads-following-a-symlinked-leaf (a lower-severity confusion vector, not a write-escape) are a noted follow-up.

### DIFF
--- a/src/compiler/index.ts
+++ b/src/compiler/index.ts
@@ -858,7 +858,7 @@ async function generateMergedPage(
   }
 
   const pagePath = path.join(root, CONCEPTS_DIR, `${entry.slug}.md`);
-  const error = await writePageIfValid(pagePath, fullPage, entry.concept.concept);
+  const error = await writePageIfValid(root, pagePath, fullPage, entry.concept.concept);
   if (!error) await deleteCandidateBySlug(root, entry.slug);
   return { error: error ?? undefined, wrotePage: !error };
 }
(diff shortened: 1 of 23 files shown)


ethanj:

PR #137, opened 2026-06-22
feat(events): 4b PR1 — append-only hash-chained event store + audit on lifecycle/relation writes

ethanj:
First of the 4b stack (do not merge yet). Closes the audit gap that lifecycle transitions and relation writes were UNAUDITED (a spec requirement: "lifecycle transitions create audit events").

- New append-only `wiki/graph/events.jsonl` event store: `EventRecord` (`evt_<ULID>`, type, origin, payload, timestamp) with a per-record checksum, header schemaVersion (reader fails closed when newer), torn-trailing-line tolerance, and a **hash chain** (each record's `prevHash` = the prior record's canonical digest) whose head is anchored outside the log in a sealed `.llmwiki/events.head` — so an edited, reordered, deleted, or truncated history is detectable.
- Lifecycle transitions and relation create/update now emit a chained audit event under the same lock as the mutation.
- The relation store's hardened confinement/no-follow primitives were generalized into a shared `src/utils/jsonl-store.ts` that both stores use, so the event store inherits all the symlink/regular-file/size-cap hardening.

A default project makes no transitions/relations → no event store → default profile byte-identical; full suite green (2313).

### DIFF
--- a/src/events/event-digest.ts
+++ b/src/events/event-digest.ts
@@ -1,0 +1,49 @@
+/**
+ * @file src/events/event-digest.ts
+ * @description The chain digest for the event store. Mirrors the relation
+ * digest's canonicalize-then-sha256 pattern: a SHA-256 over the RFC 8785 (JCS)
+ * canonicalization of an event's CONTENT fields (`{id,type,origin,actor,payload,
+ * decision,at}`), so equal content yields an equal hash regardless of key order.
+ *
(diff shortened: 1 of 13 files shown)


ethanj:
Landed on feature/configurable-profiles as squash 27fa3f7 (4b — audit event store + typed-page/relation read integration). Closing the stack; the change is on the integration branch.

PR #138, opened 2026-06-22
feat(events): 4b PR2 — fail-closed event-chain lint + event count in status

ethanj:
Stacked on PR #137 (do not merge yet). Makes the event store's tamper-evidence actually enforced on read.

- A new `checkEventChain` lint rule reads the event store and fails closed on a broken/forked/reordered/interior-deleted chain or a head-anchor mismatch (truncation) → `event-chain-broken` error; a torn trailing line → warning; corrupt/too-new/symlinked store → fail-closed findings. Lint never throws a store error.
- Status gains an optional `eventCount`; a broken chain or unreadable store surfaces a capped problem and suppresses the count (never reports a bogus healthy total) instead of crashing.

Every event-store error type is mapped into both surfaces from the start (the recurring cross-surface lesson). Default + event-less projects emit no event fields — byte-identical; full suite green (2326).

### DIFF
--- a/src/profile/block.ts
+++ b/src/profile/block.ts
@@ -26,6 +26,8 @@ import { readRelations } from "../relations/store-read.js";
 import { validateRelationAgainstProfile } from "../relations/relation-contract.js";
 import { RelationStoreCorruptError, RelationStoreTooNewError, RelationStoreSymlinkError } from "../relations/types.js";
 import type { RelationRef } from "../relations/types.js";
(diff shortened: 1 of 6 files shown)


ethanj:
Landed on feature/configurable-profiles as squash 27fa3f7 (4b — audit event store + typed-page/relation read integration). Closing the stack; the change is on the integration branch.

PR #140, opened 2026-06-22
Add graph health to eval harness

dohu012:
Builds on the existing viewer graph path (collectViewerPages +                                     
  buildGraphData) as discussed in #78.                                                               
                                                                                                     
  - unreferenced pages (indegree=0), weakly connected components,                                    
    average indegree (real-to-real only), hub pages (top 5 by                                        
    total degree, tie-broken by slug), dangling targets                                              
  - graphHealth is optional on EvalReport                                                            
  - 8 tests, tsc clean

### DIFF
--- a/src/eval/graph-health.ts
+++ b/src/eval/graph-health.ts
@@ -76,21 +76,13 @@ function countComponents(nodes: GraphNode[], edges: GraphData["edges"], realIds:
   return components;
 }
 
-/** Compute top dangling targets by indegree from real pages. */
-function topDanglingTargets(
-  nodes: GraphNode[], edges: GraphData["edges"], realIds: Set<PageId>,
-): GraphHealthResult["topDangling"] {
-  const counts = new Map<string, number>();
-  for (const e of edges) {
-    const target = nodes.find((n) => n.id === e.target);
-    if (target?.isDangling && realIds.has(e.source)) {
-      counts.set(target.title, (counts.get(target.title) ?? 0) + 1);
-    }
-  }
-  return [...counts.entries()]
-    .sort((a, b) => b[1] - a[1] || a[0].localeCompare(b[0]))
+/** Top dangling targets by indegree, using ghost nodes' pre-computed degrees. */
+function topDanglingTargets(graph: GraphData): GraphHealthResult["topDangling"] {
+  return graph.nodes
+    .filter((n) => n.isDangling)
+    .sort((a, b) => b.degree - a.degree || a.id.localeCompare(b.id))
     .slice(0, MAX_TOP_DANGLING)
-    .map(([title, count]) => ({ title, referenceCount: count }));
+    .map((n) => ({ id: n.id, title: n.title, referenceCount: n.degree }));
 }
 
 /**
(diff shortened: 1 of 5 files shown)


ethanj (changes requested):
Really nice work — this is the cleanest scope-adherence of the three eval slices. It builds on `collectViewerPages()` + `buildGraphData()` exactly as discussed (no second wikilink parser), uses the "unreferenced" naming, computes weakly-connected components over real pages, hubs by total degree tie-broken by slug, surfaces dangling targets, keeps `graphHealth` optional for back-compat, and leaves thresholds and the health score untouched. Gates are green locally (tsc, build, 1792 tests, fallow clean) and I verified the metrics on a hand-built graph — they're correct, and the terminal output holds the frame width.

Two things I'd like to fix before merge, both about **identity in the JSON** rather than the math:

**1. Pages are identified by bare slug, which is ambiguous.** `hubPages[].slug` and `unreferencedPages[]` (`src/eval/graph-health.ts:123`, `:134`) emit `n.slug`, but llmwiki allows `concepts/shared` and `queries/shared` as distinct pages — that's the whole reason `resolveBareSlug()` has a concepts-win precedence rule. With a colliding slug, a consumer of the JSON can't tell which page is the hub or which is unreferenced. Please emit the namespaced `n.id` (`concepts/foo` / `queries/foo`), or a `{ id, slug, directory }` shape if you want to keep a display slug. This also makes graph health consistent with `pageHealthDistribution`, which already keys pages by `dir/basename`. A regression test with a concept and a query sharing a slug would lock it in.

**2. `topDangling` aggregates by display title.** `src/eval/graph-health.ts:83` keys the counts by `target.title`, so two different missing targets that share an alias/display (e.g. `[[missing-a|TODO]]` and `[[missing-b|TODO]]`) collapse into one row, and the JSON has no id to repair it. Group on the ghost node instead (`target.id`) and return `{ id, title, referenceCount }`.

While you're in `topDangling`, this is also a chance to simplify: the function currently does `nodes.find(n => n.id === e.target)` per edge (O(edges × nodes)), but `buildGraphData` already stamps each ghost node's `degree` with its indegree (`buildGhostNodes` → `degree: inDegreeMap.get(target)`). So the whole thing collapses to a sort over the dangling nodes and reuses data the snapshot already computed:

```ts
const topDangling = graph.nodes
  .filter(n => n.isDangling)
  .sort((a, b) => b.degree - a.degree || a.id.localeCompare(b.id))
  .slice(0, MAX_TOP_DANGLING)
  .map(n => ({ id: n.id, title: n.title, referenceCount: n.degree }));
```

That fixes the identity issue and removes the only O(N×E) path — and it's truer to the file's own "one source of truth / all metrics derived from the snapshot" docstring.

**Minor (optional):** `hubPages` computes outdegree per node with `graph.edges.filter(e => e.source === n.id …)`, which is O(nodes × edges). Indegree already goes through the clean `buildRealIndegrees` map; a symmetric one-pass `buildRealOutdegrees` would make both O(edges). Low urgency since eval isn't a hot path, but it's the obvious shape.

**Docs:** This adds a visible terminal + JSON metric, but the eval docs don't mention it yet — `docs/cli/lint-eval.mdx:73` (the fast-suite table) and the "What eval measures" section have no graph-health entry, and `README.md:44` omits it from the eval-harness summary. Fine to fold in here or as an immediate follow-up (same shape as the page-health docs update).

**Nits, take or leave:**
- `formatGraphHealth` uses `function(h){…}` while the rest of `report.ts` uses arrow functions.
- Stray blank lines in `src/eval/graph-health.ts` (22–24) and `src/eval/types.ts`.
- `avgIndegree` is stored at two decimals (`0.83`) but rendered at one (`0.8`).
- The tie-break test wraps its assertion in `if (alphaIdx >= 0 && betaIdx >= 0)`, so it would pass vacuously if both hubs were ever dropped.

Once the two identity issues are sorted I think this is good to merge — the shape and the reuse are exactly right.

dohu012:
Both identity issues fixed — hubPages and unreferencedPages now emit namespaced n.id, and topDangling uses ghost nodes' pre-computed degree keyed by n.id. Regression test covers same-slug collision and dangling title dedup in one scenario. realOutdegrees map added for O(edges) hub computation. Build clean, CI green.

ethanj:
Thanks for the quick follow-up — the two issues are fixed cleanly.
Great work! Looking forward to other good ideas from you.
FYI I have been working on a significant upgrade of the abstraction, which will hopefully be ready soon.

Issue #39, opened 2026-04-27
prompt blowup — mergeExtractions concatenates all source content for shared concepts without bounds


lllcccwww:
## Summary

When multiple source documents contribute to the same concept, the page
generation step concatenates their **full content** into a single prompt,
with no truncation, no relevance filtering, and no map-reduce. The prompt
size grows linearly with the number of sources sharing a concept, which
makes `compile` fail on realistic corpora — and **incremental updates
re-trigger the same blowup on already-shared concepts**.

The failure is consistently observable as a context-length error from the
LLM provider (e.g. `prompt is too long`, `context window exceeds limit`).
It is naturally hit whenever the same concept is discussed across many
overlapping documents.

## Reproduction

1. Create N source documents (N ≥ 5) that all discuss the same topic.
   Each document a few thousand characters of normal prose is enough.
2. Run `llmwiki ingest` on each, then `llmwiki compile`.
3. When `compile` reaches the page-generation step for the shared topic,
   the request exceeds the LLM provider's context window and fails.

The threshold N depends on the model's context size, but the prompt size
is linear in the number of contributing sources, so any model can be
saturated by adding more documents that touch a popular concept.

## Updates

The same failure mode is also reachable during incremental updates: adding
a new source that touches an already-established concept appears to be able
to re-trigger the blowup on existing pages.

ethanj:
Thanks for the detailed repro and the call-out about the incremental-update path — that was the part that would have been easy to miss. Defensive fix in #47: when many sources contribute to the same concept, `mergeExtractions` now applies a fair-share character budget (default 200,000 chars, override via `LLMWIKI_PROMPT_BUDGET_CHARS`) and emits a clear warning when truncation kicks in. Typical workloads stay byte-identical; popular concepts no longer crash compile. Smarter retrieval-driven slicing (rank by relevance instead of fair-share) is a follow-up — wanted to ship the crash-prevention first.

Issue #11, opened 2026-04-18
Ollama/OpenAI client path appears to inherit a 10-minute default timeout during long compile run


BenGSt:
I’m using llm-wiki-compiler with:
	•	LLMWIKI_PROVIDER=ollama
	•	OLLAMA_HOST=http://host.docker.internal:11434/v1
	•	LLMWIKI_MODEL=gemma4:e4b

and I’m seeing long compile requests fail at almost exactly 10 minutes.

What happens

During compilation, some model calls succeed, but longer ones fail right around the 10-minute mark. On the Ollama side, I see logs like:
	•	POST "/v1/chat/completions" returns after ~10m0s
	•	then Ollama logs: aborting completion request due to client closing the connection

This makes it look like the client times out first, and then Ollama aborts because the socket closes.

Why I think this is happening

From the installed package, the Ollama provider inherits from OpenAIProvider, and OpenAIProvider constructs the JS client like this:
	•	constructor(model, baseURL, apiKey)
	•	this.client = new OpenAI({ apiKey: resolvedKey, ...(baseURL ? { baseURL } : {}) })
	•	requests then go through this.client.chat.completions.create(...)
	•	OllamaProvider calls super(model, baseURL, "ollama")

I do not see a timeout passed into the new OpenAI(...) client construction, so this appears to inherit the OpenAI SDK default timeout.  ￼

Suggested fix

Please make the client timeout configurable, for example:
	•	add timeout when constructing new OpenAI(...)
	•	optionally support an env var such as LLMWIKI_TIMEOUT_MS

For example:

this.client = new OpenAI({
  apiKey: resolvedKey,
  timeout: 30 * 60 * 1000,
  ...(baseURL ? { baseURL } : {}),
});

Why this matters

With local Ollama models, especially on larger inputs or slower hardware, some compile steps can legitimately take more than 10 minutes. Right now that seems to make compile brittle for local-model users.

Environment
	•	llm-wiki-compiler installed globally via npm in Docker
	•	Node 20 container
	•	Ollama running on host macOS
	•	Ollama exposed to container at /v1
	•	model: gemma4:e4b

Nice-to-have

It would be great if this were documented for Ollama/local-model users, since long-running local completions are more likely than with hosted APIs.

ethanj:
[...]

PR #145, opened 2026-07-04
fix(ollama): use native /api/chat format for structured tool calls

JulianKominovic:
## Summary

Fixes concept extraction (and other structured tool calls) failing silently when using the Ollama provider with local models such as `llama3.1:latest`.

`OllamaProvider` previously inherited `OpenAIProvider.toolCall()`, which routes through the OpenAI-compatible `/v1/chat/completions` endpoint with `tool_choice: "required"`. With Ollama, that path sometimes returns `function.arguments` where array fields like `concepts` are double-encoded as JSON strings instead of native arrays:

```json
{ "concepts": "[{\"concept\":\"...\",\"summary\":\"...\",\"is_new\":true}]" }
```

`parseConcepts` then hits `concepts.filter is not a function`, catches the error, and returns `[]`. The compile pipeline surfaces `no concepts — will retry`, writes zero wiki pages, and still reports the source as "compiled" — a misleading partial success.

### 1. `OllamaProvider.toolCall` — native structured output (primary)

Override `toolCall()` to call Ollama's native `POST /api/chat` with `format: tool.input_schema` (JSON-schema structured output). This applies to **all** structured tool calls in the pipeline: concept extraction, rule extraction, query page selection, and eval judge.

- Embeddings and other OpenAI-compatible calls continue to use `OLLAMA_HOST` / `OLLAMA_EMBEDDINGS_HOST` via the existing `/v1` client — unchanged.
- The native base URL is derived from `OLLAMA_HOST` by stripping the trailing `/v1` suffix while **preserving any reverse-proxy path prefix** (e.g. `https://proxy/ollama/v1` → `https://proxy/ollama/api/chat`).
- Uses `fetch` directly (no new `ollama` npm dependency).
- Respects `OLLAMA_TIMEOUT_MS` / `LLMWIKI_REQUEST_TIMEOUT_MS` via `AbortController`.

### 2. `parseConcepts` — defensive coercion (secondary)

Coerce `concepts` when it arrives as a JSON string, and return `[]` cleanly when it is not an array after coercion. This does not fix the root cause but protects against the same quirk from other providers/models.

## Error handling

Invalid `OLLAMA_HOST` values now throw at provider construction instead of silently falling back to localhost. Failed `/api/chat` HTTP responses and empty message content also throw with actionable error messages, rather than returning empty output that the pipeline misreads as "no concepts."

## Differences from initial investigation notes

The implementation follows the same two-change strategy documented in local investigation notes, with refinements discovered during implementation:

| Topic | Initial notes | Final implementation |
|---|---|---|
| Native host resolution | Strip to `protocol//host` only | Strip `/v1` from pathname but **keep path prefixes** for reverse proxies |
| Invalid `OLLAMA_HOST` | Silent fallback to localhost | **Throw** `Invalid OLLAMA_HOST URL: …` at construction |
| Request body | Included `think: false` | Omitted — not required for this use case |
| `parseConcepts` tests | Suggested new `test/compiler/prompts.test.ts` | Added to existing `test/confidence-metadata.test.ts` |
| `OllamaProvider` tests | Basic `/api/chat` + `format` mock | Extended: proxy path prefix, embeddings host isolation, HTTP errors, empty content, timeout abort |
| Docs | Note in providers page | Updated `providers.mdx` and `environment-variables.mdx` |

## Test plan

- [x] Unit: `parseConcepts` coerces string-encoded `concepts` array
- [x] Unit: `parseConcepts` returns `[]` when string-encoded `concepts` is invalid JSON
- [x] Unit: `OllamaProvider.toolCall` hits `/api/chat` with `format` schema (not OpenAI `tool_calls`)
- [x] Unit: reverse-proxy path prefix preserved in native URL
- [x] Unit: `OLLAMA_EMBEDDINGS_HOST` not used for native tool calls
- [x] Unit: invalid `OLLAMA_HOST`, HTTP errors, empty content, and timeout abort throw correctly
- [x] `npm test`, `npm run build`, `npx tsc --noEmit`
- [x] Manual: `LLMWIKI_PROVIDER=ollama llmwiki compile` against a local Ollama instance produces concept pages (not `no concepts — will retry`)

## Notes for operators behind a reverse proxy

`OLLAMA_HOST` must still include `/v1` for the OpenAI-compatible client. The proxy must route both:

- `/your-prefix/v1/*` → Ollama OpenAI-compatible API (embeddings)
- `/your-prefix/api/*` → Ollama native REST API (`/api/chat`)

## Related

Distinct from timeout ([#11](https://github.com/atomicstrata/llm-wiki-compiler/issues/11)) and context-window blowup ([#39](https://github.com/atomicstrata/llm-wiki-compiler/issues/39)) issues — this is **malformed structured output** on the OpenAI-compatible path, not a timeout or context problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **Ollama structured tool calls** — `OllamaProvider.toolCall` now uses Ollama's native `POST /api/chat` with JSON-schema `format` (base URL derived from `OLLAMA_HOST`, preserving reverse-proxy path prefixes) instead of the OpenAI-compatible `tool_calls` path, which could return double-encoded `concepts` arrays and silently produce zero wiki pages. Affects all structured tool calls (concept extraction, rule extraction, query page selection, eval judge). Invalid `OLLAMA_HOST` values and empty `/api/chat` responses now throw instead of silently falling back or returning empty output. `parseConcepts` also coerces string-encoded `concepts` arrays as a defensive fallback. Embeddings continue to use `OLLAMA_HOST` / `OLLAMA_EMBEDDINGS_HOST` via the OpenAI-compatible `/v1` client.
+
 ## [0.11.0] - 2026-06-16
 
 Extends Open Knowledge Format support beyond the CLI: in-process SDK and MCP access to the OKF round-trip, and faithful reconstruction of an imported foreign bundle's original nested paths on re-export.
--- a/docs/configuration/environment-variables.mdx
+++ b/docs/configuration/environment-variables.mdx
@@ -46,7 +46,7 @@ Either `ANTHROPIC_API_KEY` or `ANTHROPIC_AUTH_TOKEN` satisfies authentication - 
 
 | Variable | Required | Description |
 |---|---|---|
-| `OLLAMA_HOST` | Yes (for `ollama` provider) | Ollama host URL for chat and tool calls. **Must include `/v1`** |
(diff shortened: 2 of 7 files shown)


ethanj:
I pushed a small test-only cleanup commit to clear the fallow duplication gate. No provider logic changed.

ethanj:
Thanks for taking this on, and welcome as a first-time contributor to llmwiki.

This is a really helpful fix. Ollama support matters a lot for people running llmwiki locally, and this clears up one of those frustrating cases where things looked like they “worked” but quietly produced no pages. I also appreciated the care you put into the explanation, docs, proxy-path handling, and tests.

I pushed one small test-only cleanup commit to clear the fallow duplication check.

PR #166, opened 2026-07-23
fix(citations): align line-list parsing

typevolant:
The compile-time normalizer emits comma-separated line lists like `file.md:12,15,20` and mixed lists like `file.md:1-5,12`, but nearby citation consumers re-derived that grammar differently.

This fix:
- Allows hyphen-range references in comma series
- Shares citation line-range parsing with malformed validation and bounds checks
- Checks every line-list range for out-of-bounds citations
- Uses the shared citation splitter in viewer warnings
- Adds regression coverage for normalizer round-trips, non-leading ranges, list bounds checks, and viewer warnings

### DIFF
--- a/src/linter/rules-citations.ts
+++ b/src/linter/rules-citations.ts
@@ -4,16 +4,18 @@
  * This family validates the `^[file.md]` citation markers that anchor wiki
  * prose to its sources: broken citations (source file missing or claim-level
  * line span out of bounds) and malformed claim citations (entries that don't
- * parse against the documented `file.md` / `file.md:N-N` / `file.md#LN-LN`
- * grammar). Each on-disk walker has a pure per-page variant so the in-memory
- * candidate-lint path (`compile --review`) surfaces the same findings before a
- * reviewer approves a candidate. Re-exported through {@link file://./rules.ts}.
+ * parse against the documented `file.md` / `file.md:N-N` / `file.md:N,N-M`
+ * / `file.md#LN-LN` grammar). Each on-disk walker has a pure per-page variant
+ * so the in-memory candidate-lint path (`compile --review`) surfaces the same
+ * findings before a reviewer approves a candidate. Re-exported through
+ * {@link file://./rules.ts}.
  */
 
 import { existsSync } from "fs";
 import path from "path";
 import {
   isMalformedCitationEntry,
+  parseCitationLineRanges,
   parseFrontmatter,
   safeReadFile,
   splitCitationMarker,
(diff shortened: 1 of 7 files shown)


ethanj (changes requested):
### Review

Thanks for this — and welcome! This is a good catch: the normalizer emitting a shape its own validator rejected is exactly the kind of silent grammar drift that's easy to miss, and routing both the parser and the validator through a shared `parseLineToken`/`isValidLineRange` primitive is the right instinct. The round-trip test suite in `test/citation-normalize.test.ts` is a nice touch too; that's the control that keeps this from re-drifting.

I dug into the surrounding call graph and found a few places where the widened grammar doesn't survive contact with its neighbors. Flagging them below with repros. CI is green and the local suite (4,326 tests) passes — these all sit in gaps the current tests don't reach, so no fault in the signal you were working from.

**1. Blocking — a non-leading range token is silently dropped**

`COLON_MULTILINE_PATTERN` now accepts a hyphen range in any comma position, but `splitCitationMarker` is unchanged and only protects a comma followed by a *plain* digit token. `^[source.md:1, 12-15]` splits into `source.md:1` and `12-15`; extraction treats `12-15` as a filename and multi-source normalization drops it, so the citation quietly becomes `^[source.md:1]`. `isMalformedCitationEntry` doesn't catch the orphan either — it has no `:` or `#`, so it early-returns as well-formed.

https://github.com/atomicstrata/llm-wiki-compiler/blob/f065a26f0db4b1a9b03bd6936ef861561d1da699/src/utils/markdown.ts#L181-L184

https://github.com/atomicstrata/llm-wiki-compiler/blob/f065a26f0db4b1a9b03bd6936ef861561d1da699/src/utils/markdown.ts#L30-L32

Suggested fix: give the split lookahead the same per-token grammar the pattern uses (`\d+(?:-\d+)?`), so the two agree by construction. The existing tests all put the range first (`1-5, 12`), which is why this slips through — a case with the range in second position would pin it.

**2. Blocking — newly accepted lists bypass bounds checking entirely**

Accepting multi-token lists in `isMalformedCitationEntry` removes the only finding these entries used to produce. `checkPageBrokenCitations` uses the older single-range `COLON_SPAN_PATTERN`, which returns `null` for `source.md:1,2,12`, and a `null` range means an early `continue` — no bounds check at all. Against a three-line source, `^[source.md:1,2,12]` now yields zero malformed *and* zero broken findings, where before it was at least flagged malformed.

https://github.com/atomicstrata/llm-wiki-compiler/blob/f065a26f0db4b1a9b03bd6936ef861561d1da699/src/linter/rules-citations.ts#L29-L31

https://github.com/atomicstrata/llm-wiki-compiler/blob/f065a26f0db4b1a9b03bd6936ef861561d1da699/src/linter/rules-citations.ts#L152-L156

Related: `^[source.md:12,1]` *does* match `COLON_SPAN_PATTERN`, producing `start=12, end=1`, and only `range.end` is bounds-checked — so it passes too. Having `parseLineRange` consume the same shared token parser (and check every token, not just the last) would close both.

**3. Non-blocking — the viewer path still warns on valid normalizer output**

`appendCitationWarningsForMarker` splits on `raw.split(",")` instead of the shared `splitCitationMarker`, so a valid `^[source.md:1, 12]` still surfaces `unresolved_citation: Source not found: 12` in the viewer. This one predates your change, but it's squarely inside the invariant the PR asserts ("normalizer output is always accepted"), and the new round-trip tests only exercise the linter. Worth either fixing here or noting as a follow-up — your call.

https://github.com/atomicstrata/llm-wiki-compiler/blob/f065a26f0db4b1a9b03bd6936ef861561d1da699/src/viewer/snapshot.ts#L242-L246

**4. Minor — module docblock and error message disagree with the new grammar**

The message on line 210 gained a fourth form, but the module header two screens up still documents only `file.md` / `file.md:N-N` / `file.md#LN-LN`. The file now states two grammars for itself; the repo's contributor guidelines ask that comments stay current with code changes. The new message text also under-describes what's accepted — `file.md:N,N,…` omits the hyphen-range-in-list form this PR legalizes.

https://github.com/atomicstrata/llm-wiki-compiler/blob/f065a26f0db4b1a9b03bd6936ef861561d1da699/src/linter/rules-citations.ts#L3-L11

https://github.com/atomicstrata/llm-wiki-compiler/blob/f065a26f0db4b1a9b03bd6936ef861561d1da699/src/linter/rules-citations.ts#L209-L211

The through-line on 1–3 is that `COLON_MULTILINE_PATTERN` now has three sibling consumers that each re-derive the token grammar their own way. Having all of them go through the shared token parser you introduced would make the drift structurally impossible rather than test-dependent — and that would make this a stronger fix than the one you set out to write. Happy to look again whenever you've pushed.

typevolant:
Makes sense. Added those updates. =)

ethanj (approved):
Approving — nice work, and a strong first contribution.

You didn't just patch the four things I flagged; you went after the root cause instead. Routing the linter, the extractor, and the validator through one shared parser means they can no longer disagree about what a line token is, which is what makes this class of drift structurally impossible rather than something we keep writing tests against. Landing that as a net-negative diff while adding coverage is a good trade.

Thanks for the care on the tests, and for reaching for the shared primitive rather than the symptoms.

PR #55, opened 2026-05-10
feat: added GitHub Copilot provider

cadamsdev:
Adds `copilot` as a new LLM provider for users with a GitHub Copilot subscription.
 
 ### Changes
 
 1. **`src/providers/copilot.ts`** — new provider extending `OpenAIProvider` targeting 
`https://api.githubcopilot.com`
 2. **`src/providers/openai.ts`** — added `tool_choice: "required"` to `toolCall()` so Claude models don't skip 
tool calls and return empty results
 3. **`src/utils/provider.ts`, `src/cli.ts`, `src/mcp/provider-check.ts`** — wired `copilot` into all three 
provider registries
 4. **`test/provider-copilot.test.ts`** — 7 unit tests
 5. **`README.md`** — setup docs
 
 ### Setup
 
 ```bash
 gh auth refresh --scopes copilot
 export LLMWIKI_PROVIDER=copilot
 export GITHUB_TOKEN=$(gh auth token)
 export LLMWIKI_MODEL=claude-sonnet-4.6  # optional; gpt-4o is the default
```

 Note: The Copilot API has no embeddings endpoint — compile and query work normally but semantic chunked retrieval
 is unavailable.

Preview
Screenshot 2026-05-10 at 11 47 13 AM

Screenshot 2026-05-10 at 11 41 34 AM

PR #167, opened 2026-07-24
Add Atlas Cloud LLM provider

binyangzhu000-sudo:
## Summary
- add an Atlas Cloud OpenAI-compatible LLM provider with atlascloud/atlas-cloud/atlas aliases
- support ATLASCLOUD_API_KEY / ATLAS_CLOUD_API_KEY and Atlas base URL env aliases while defaulting to https://api.atlascloud.ai/v1
- keep embeddings fail-closed until an Atlas-compatible embedding model is verified, matching existing provider safety patterns

## Validation
- npm ci --ignore-scripts
- npx vitest run test/provider-atlascloud.test.ts test/provider-factory.test.ts test/utils/provider-guard.test.ts
- npx tsc --noEmit
- npm run build
- npm run release:check-docs
- npm run fallow:ci
- npm test (606 files passed, 1 skipped; 4329 tests passed, 3 skipped)
- Atlas live /api/v1/models returned 437 models / 375 visible and confirmed qwen/qwen3.5-flash and deepseek-ai/deepseek-v4-pro as visible Text models

## Notes
- No README/docs/logo changes; no sponsor, credits, or partner copy.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -345,7 +345,7 @@ program
   .option("--no-open", "Skip the viewer handoff after a successful compile")
   .option(
     "--provider <name>",
-    "Override LLMWIKI_PROVIDER for this run only (e.g. anthropic, openai, ollama)",
+    "Override LLMWIKI_PROVIDER for this run only (e.g. anthropic, openai, ollama, atlascloud)",
   )
   .option(
     "--lang <code>",
--- a/src/eval/citation-support.ts
+++ b/src/eval/citation-support.ts
@@ -17,7 +17,12 @@ import path from "path";
 import { collectAllPages } from "../linter/rules.js";
 import { parseFrontmatter, extractClaimCitations, splitProseParagraphs } from "../utils/markdown.js";
 import { callClaude } from "../utils/llm.js";
-import { SOURCES_DIR, DEFAULT_PROVIDER, PROVIDER_MODELS } from "../utils/constants.js";
+import {
+  SOURCES_DIR,
+  DEFAULT_PROVIDER,
+  PROVIDER_MODELS,
+  normalizeProviderName,
+} from "../utils/constants.js";
 import { resolveSourceFile } from "./source-path.js";
 import type { LLMTool } from "../utils/provider.js";
 import type { CitationJudgement, CitationSupportResult } from "./types.js";
@@ -212,7 +217,7 @@ async function appendCachedJudgement(root: string, judgement: CitationJudgement)
 
 /** Resolve the current model identifier for recording in judgements. */
 function resolveModel(): string {
-  const provider = process.env.LLMWIKI_PROVIDER ?? DEFAULT_PROVIDER;
+  const provider = normalizeProviderName(process.env.LLMWIKI_PROVIDER ?? DEFAULT_PROVIDER);
   return process.env.LLMWIKI_MODEL ?? PROVIDER_MODELS[provider] ?? provider;
 }
 
--- a/src/providers/atlascloud.ts
+++ b/src/providers/atlascloud.ts
@@ -1,0 +1,52 @@
+/**
+ * Atlas Cloud LLM provider implementation.
+ *
+ * Atlas Cloud exposes an OpenAI-compatible chat completions API. Embeddings are
+ * not wired until a compatible embedding model is verified for llmwiki.
+ */
+
+import { OpenAIProvider } from "./openai.js";
+import {
+  ATLASCLOUD_API_KEY_ENV_VARS,
+  ATLASCLOUD_BASE_URL,
+  ATLASCLOUD_BASE_URL_ENV_VARS,
+} from "../utils/constants.js";
+
+function readFirstEnv(names: readonly string[]): string | undefined {
+  for (const name of names) {
+    const value = process.env[name]?.trim();
+    if (value) return value;
+  }
+  return undefined;
+}
+
+/** Resolve Atlas Cloud API key from the supported env-var aliases. */
+export function resolveAtlasCloudApiKeyFromEnv(): string | undefined {
+  return readFirstEnv(ATLASCLOUD_API_KEY_ENV_VARS);
+}
+
+/** Resolve Atlas Cloud OpenAI-compatible base URL from env, or use the default. */
+export function resolveAtlasCloudBaseURLFromEnv(): string {
+  return readFirstEnv(ATLASCLOUD_BASE_URL_ENV_VARS) ?? ATLASCLOUD_BASE_URL;
+}
+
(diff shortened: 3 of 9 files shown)


ethanj (changes requested):
### Review

Thanks for this, and welcome — this is a nicely built first contribution. You clearly went and read how the existing providers work before writing anything, and it shows: `AtlasCloudProvider` mirrors the `minimax.ts`/`copilot.ts` shape closely, the fail-closed `embed()`/`embedBatch()` overrides follow the established convention rather than silently inheriting OpenAI's, and the alias handling is threaded through *every* site that reads a provider string — including `resolveModel()` in `src/eval/citation-support.ts`, which was reading it raw before. That last one is a genuine drive-by fix. The test coverage for alias routing and missing-key behavior is good, and I appreciated the clean env-var precedence with no fallback to another provider's key.

A few things to sort out before this lands. The first is the one that matters.

**1. Blocking — the default model isn't verified against the tool-calling contract compile depends on**

Atlas inherits `OpenAIProvider`, which sends `tools` and `tool_choice: "required"` on completions:

https://github.com/atomicstrata/llm-wiki-compiler/blob/72305a3c9bfa4bc580fdfd30e36199063a602c36/src/providers/openai.ts#L143-L146

Concept extraction and several other primary compile paths depend on getting tool arguments back — a model that ignores `tools` or can't satisfy `tool_choice: "required"` doesn't degrade gracefully here, it fails the workflow. The default is set to `qwen/qwen3.5-flash`:

https://github.com/atomicstrata/llm-wiki-compiler/blob/72305a3c9bfa4bc580fdfd30e36199063a602c36/src/utils/constants.ts#L97-L99

The new tests confirm the provider constructs and that the model string resolves, but none of them exercises a real structured completion, so the tool contract is currently unvalidated for this default. Could you confirm that model returns tool calls against Atlas's endpoint — ideally with a contract test that asserts tool arguments come back, not just that the model id resolves? If it turns out it doesn't support tool calling, the fix is to pick a tool-capable default or add an Atlas-specific structured-output path. Happy to be wrong here; I just couldn't verify it from the published model page either way, and it's the one thing that would make the provider look installed but not actually work.

**2. Blocking — user-facing provider and env vars are missing from the docs**

`docs/AGENTS.md:18` requires that any user-facing feature change update the Mintlify docs in the same PR, and line 19 names **providers** and **environment variables** explicitly. Three surfaces need updating:

- `docs/configuration/environment-variables.mdx:17` — the `LLMWIKI_PROVIDER` table enumerates the valid values and omits `atlascloud`, which makes it factually wrong rather than merely incomplete
- `docs/configuration/providers.mdx` — needs an Atlas Cloud setup tab, and it's worth calling out the embeddings limitation there so users know semantic search needs a different provider
- `README.md` — provider list and credentials table

https://github.com/atomicstrata/llm-wiki-compiler/blob/72305a3c9bfa4bc580fdfd30e36199063a602c36/src/utils/constants.ts#L104-L114

For precedent, the two most recent provider additions (#55 and #81) both did this in the same PR.

**3. Worth fixing — semantic search hard-fails instead of degrading**

`EMBEDDING_MODELS` has no `atlascloud` entry, so model resolution falls through to the Anthropic default:

https://github.com/atomicstrata/llm-wiki-compiler/blob/72305a3c9bfa4bc580fdfd30e36199063a602c36/src/utils/embeddings-store.ts#L302-L307

The effect: with an existing voyage-3-lite store on disk, the freshness gate resolves Atlas to `voyage-3-lite`, decides the store is current, and starts the semantic path — which then hits your `embed()` override and throws, instead of falling back to lexical search.

https://github.com/atomicstrata/llm-wiki-compiler/blob/72305a3c9bfa4bc580fdfd30e36199063a602c36/src/providers/atlascloud.ts#L38-L50

Important caveat, and this is not on you: MiniMax and Copilot are in exactly the same position — absent from `EMBEDDING_MODELS`, throwing on `embed()`. So this is a pre-existing gap that your provider joins rather than one you introduced, and it may well be better fixed as a separate change covering all three (either an explicit "no embeddings" marker that makes the store gate degrade immediately, or catching embedding-unavailable at the semantic-selection call site). Your call whether to take it here or file it — I don't want to widen the scope of a first PR for something two existing providers already do.

**4. Minor — provider list now lives in three places**

The unknown-provider error previously read from `Object.keys(PROVIDER_KEY_VARS)`, the same object doing the credential lookup, so the advertised list couldn't drift from the enforced one. It now reads `SUPPORTED_PROVIDER_INPUTS`, leaving three separately-maintained lists (`constants.ts`, `SUPPORTED_PROVIDERS` in `provider.ts`, `PROVIDER_KEY_VARS` in `provider-guard.ts`). All three agree today so nothing is broken — but the next provider that updates two of three gets a guard that rejects what the factory can build. Worth collapsing to one source while it's cheap.

**5. Minor — the file header you edited is still stale**

https://github.com/atomicstrata/llm-wiki-compiler/blob/72305a3c9bfa4bc580fdfd30e36199063a602c36/src/utils/provider.ts#L5-L7

You added Atlas Cloud to this list, but it still omits the Copilot and Agent-SDK providers, both implemented in that same file. Easy to finish off while you're in there.

Item 1 is the real gate; 2 is a repo rule; 3 is genuinely optional for this PR. Nice work on the parts that are done — happy to talk any of it through if something's unclear or if you'd rather split some of it into a follow-up.

ethanj:
Following up on this after a month, and rather than leave it sitting I have picked it up and finished it in #190. You are credited as co-author there: the provider, the alias handling and the original tests are your work, and they rebased onto current `main` with no changes needed.

Two things I changed, and one is worth knowing about.

The default model. I was able to check the tool-calling question I raised, and the answer was no. Atlas Cloud's catalogue marks tool support per model, and `qwen/qwen3.5-flash` carries no `supported_features` entry at all, while 35 of its 136 models advertise `["json_mode","structured_outputs","tools"]`. Since `compile` extracts concepts through a tool call with `tool_choice: "required"`, that default would have failed on the first request rather than degrading. It is now `qwen/qwen3.5-35b-a3b`, the smallest model in the catalogue that does advertise tools, which keeps the cheap-and-fast choice you were clearly going for. Nothing you could have known without the catalogue in front of you, and the rest of the provider was built correctly around it.

Docs, which were the other blocking item, plus the provider-list consolidation I mentioned in point 4.

Thank you for this. The alias threading through `resolveModel()` in `citation-support.ts` was a real drive-by fix that had nothing to do with your provider, and it stayed in. If you want to pick up the embeddings gap from point 3 as a separate change covering MiniMax and Copilot too, that would be a genuinely useful follow-up and I would review it quickly.

Closing this in favour of #190.

Issue #144, opened 2026-07-03
Is there a supported way to customize the compile prompt or provide a SOUL.md-style instruction file?


carmilso:
Hi! First of all, thanks for building `llmwiki` — the compile-time wiki pattern is exactly what I was looking for.

I’m trying to understand whether there is a supported way to customize the instructions used during `llmwiki compile`, especially the prompt/instructions used for concept extraction and page generation.

My use case is that I’d like to provide project-level guidance similar to a `SOUL.md` / “system instructions” file, for example:

- preferred writing style and structure for generated wiki pages
- domain-specific terminology
- what kinds of concepts should or should not become pages
- preferred level of detail
- project-specific rules for citations, wikilinks, summaries, or page organization

I saw that `llmwiki compile` supports options such as `--review`, `--lang`, and `--concurrency`, but I couldn’t find a documented way to pass custom compile instructions or a custom prompt file.

A few questions:

1. Is there currently any supported way to customize the compile prompt without forking/editing the source?
2. Is a file such as `SOUL.md`, `CLAUDE.md`, `AGENTS.md`, or another project-level instruction file read by `llmwiki compile`?
3. Is `.llmwiki/config.json` or `.llmwiki/schema.json` intended to support this kind of prompt customization?
4. If not, would you be open to a feature like one of these?

```bash
llmwiki compile --instructions ./SOUL.md

ethanj:
Thanks for the thoughtful write-up. I think this is a genuinely useful direction.

Right now there is no supported way to customize the compile prompt without changing source code. `SOUL.md`, `CLAUDE.md`, and `AGENTS.md` are not read by `llmwiki compile`; `.llmwiki/config.json` is currently review-policy config; and `.llmwiki/schema.json` is for page kinds / seed pages rather than broad prompt customization.

I’d be very open to adding this. My preferred shape would be explicit additive instructions, not full prompt replacement, something like:

`llmwiki compile --instructions ./SOUL.md`

The instruction file could refine writing style, terminology, detail level, concept-selection preferences, and organization, while the core compiler rules would still stay authoritative: source grounding, citations, provenance, wikilinks, and valid wiki output.

A good first PR could:

- add `--instructions <path>` to `compile`;
- inject the bounded file contents into extraction/page-generation prompts as “Project instructions”;
- fail clearly on missing/oversized files;
- avoid auto-reading `CLAUDE.md` / `AGENTS.md` by default;
[...]

PR #170, opened 2026-07-30
feat(compile): add an additive system policy hook

TigerOfCountryYao:
Adds an optional additive systemPolicy compile input so host applications can append safety and publication policy without replacing the compiler prompts. Existing behavior is unchanged when omitted. This branch is stacked on #169; the intended incremental review is commit 77a43a7. Verified with TypeScript, build, targeted tests, and changed-scope checks.

### DIFF
--- a/docs/guides/sdk.mdx
+++ b/docs/guides/sdk.mdx
@@ -81,6 +81,7 @@ Both methods return `IngestResult` with `filename`, `chars`, and `truncated` fie
 
   - `options.review` - when `true`, writes generated pages to `.llmwiki/candidates/` for review instead of directly to `wiki/`. Same behavior as `llmwiki compile --review`.
   - `options.embeddings` - set to `false` to skip embedding generation and pending-embedding retries. Page generation, links, and the lexical index still run.
+  - `options.systemPolicy` - a trusted caller policy appended to both extraction and page-writing system prompts. It supplements rather than replaces the compiler's built-in instructions.
 
   Source content is sent to the configured LLM provider during compilation. Do not compile wikis containing confidential data unless the provider's data-handling policies are acceptable for that content.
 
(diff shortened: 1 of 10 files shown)


ethanj:
This shipped in #195, merged as `b22d6eb6`. You are credited in the commit body, and the credit is specific because your design is what went in rather than just the idea.

What survives from your branch:

- The additive framing, and its wording **verbatim**: `Additional system policy (follow in addition to every built-in instruction above):`. That line is what the model actually reads, and "in addition to" rather than "instead of" carries the whole contract. I did not improve on it because it was already right.
- Placing the policy after every built-in instruction and before the source material, so the full instruction block still precedes the content it describes.
- Blank treated as omitted, leaving the default prompt byte-identical.
- The `PROMPT_VERSION` bump to `v2`. You made that call unprompted, and it matches a ruling made separately on #183: the constant names the prompt implementation, so a new content-altering branch is a new generation.

What changed, and why.

**The policy is now registered as a prompt modifier.** This was the blocking gap. A policy changes what the prompt asks for without changing a source byte, and change detection classifies a source purely by the SHA-256 of its bytes. So on a settled project, compiling with Policy B after Policy A reported "Nothing to compile" and kept every page generated under the policy you had already replaced. A pending review candidate had the same problem.

You could not have known the fix: `activePromptModifiers` landed in #188, three weeks after you wrote this. Registering the policy there means it inherits the whole mechanism - invalidation on change, candidate-aware deduplication, the scoped-refresh guard, and per-page provenance. Bumping `PROMPT_VERSION` alone does not achieve it, since two different policies share a version.

**Two smaller changes.** The prompt builders read the policy from run state rather than taking it as a parameter, so their signatures are unchanged and it is not threaded through extraction, page rendering, the review pipeline and seed pages. And the modifier set records the policy's **digest**, never its text, because that set is hashed into `state.json`, stamped onto every page's frontmatter, and published through the JSON export.

On #144: this is the SDK primitive underneath that request, not a closure of it. That issue asks for a `SOUL.md`-style file, and a file or CLI surface on top of this is still open if you want it.

#169 is still open with two small asks on it, and the `finalizeWiki` signature conflict that was blocking it is resolved in #192. Thank you for this one - the prompt-construction judgement was the hard part and you got it right.

PR #171, opened 2026-07-30
fix(compile): reconcile shared pages after source deletion

TigerOfCountryYao:
Rebuilds shared pages through a fixed-point owner closure after source deletion, filters non-surviving citations during clean rebuilds, and orphans ownerless frozen slugs. Covers deletion chains, multiple deleted owners, late-discovered owners, unknown citations, ownerless pages, and survivor extraction failure. This branch is stacked on #169 and #170; the intended incremental review is commit 160e2af. Verification: 96 targeted tests, TypeScript, build, changed-scope checks, and diff checks pass.

### DIFF
--- a/docs/guides/sdk.mdx
+++ b/docs/guides/sdk.mdx
@@ -81,6 +81,7 @@ Both methods return `IngestResult` with `filename`, `chars`, and `truncated` fie
 
   - `options.review` - when `true`, writes generated pages to `.llmwiki/candidates/` for review instead of directly to `wiki/`. Same behavior as `llmwiki compile --review`.
   - `options.embeddings` - set to `false` to skip embedding generation and pending-embedding retries. Page generation, links, and the lexical index still run.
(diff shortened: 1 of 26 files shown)


ethanj:
Hi @TigerOfCountryYao did you mean for these to be draft, or do y ou want to submit them as ready for review?

ethanj:
Landed on `main` in #199 as `19631a49`, with your work credited in the merge commit.

Thank you for this. The distinction at the centre of it, that a persisted frozen slug is a reconciliation marker rather than a permanent skip, is what made the fix work, and your test suites passed against current `main` unchanged after six merges of drift.

What was added before merging: coverage for the `rm` route, which writes to that same persisted field, and fixes for four defects that came up in review. Only `160e2af7` was cherry-picked, since its parents landed separately as #198 and #195, and replaying the whole stack would have brought back implementations we had already replaced.

Closing as superseded by #199.

PR #172, opened 2026-08-03
fix(profile): compare declared paths with the POSIX separator on win32

### LINKED ISSUE #163: Windows: every 'llmwiki template init' fails — path-separator mismatch in profile directory validation
## Summary

On Windows, `llmwiki template init <template>` fails for **every** template (built-in or local) with:

```
Error: entity directory must be under 'wiki/': wiki/papers
```

so the Configurable Lifecycle Profiles feature is unusable on win32.

## Environment

- llm-wiki-compiler **1.1.0** (npm, global install)
- Windows 11 Pro, Node v24.18.0

## Repro

```powershell
mkdir probe; cd probe
llmwiki template init autosci
# Error: entity directory must be under 'wiki/': wiki/papers
```

## Root cause

In `dist/index.js` (and the copy bundled into `dist/cli.js`):

- `normalizeDeclaredDir()` normalizes declared profile directories to **posix** separators:
  ```js
  return segments.filter((s) => s !== "" && s !== ".").join(path.posix.sep); // "wiki/papers"
  ```
- but `validateEntityDirectory()` then calls `isInsideDir()`, which builds its prefix with the **platform** separator:
  ```js
  function isInsideDir(child, dir) {
    if (child === dir) return true;
    const prefix = dir.endsWith(path.sep) ? dir : dir + path.sep; // "wiki\" on win32
    return child.startsWith(prefix);                              // "wiki/papers".startsWith("wiki\") -> false
  }
  ```

So `"wiki/papers"` is checked against the prefix `"wiki\"` and every declared entity directory is rejected on Windows. The same mismatch cannot be hit on posix, which is presumably why CI doesn't catch it.

## Suggested fix

Make the containment check separator-agnostic (also safe for the realpath-confinement call sites, which pass native absolute paths for both arguments):

```js
function isInsideDir(child, dir) {
  const c = String(child).split(/[\\/]/).join("/");
  const d = String(dir).split(/[\\/]/).join("/");
  if (c === d) return true;
  return c.startsWith(d.endsWith("/") ? d : d + "/");
}
```

Locally patching `isInsideDir` exactly this way makes `template init autosci` succeed on Windows (profile validates, workflows list correctly), with existing default-profile projects unaffected.

Happy to open a PR with this change plus a win32-style regression test if useful.

PipDscvr:
Fixes #163.

@squ1ddy diagnosed this exactly — the root cause, both code excerpts, and the repro are all correct as reported. `autosci` declares `directory: "wiki/papers"`, `normalizeDeclaredDir` rejoins with `path.posix.sep`, and `isInsideDir` builds its prefix with `path.sep`, so on win32 `"wiki/papers".startsWith("wiki\\")` is false and every template fails to init.

## Two more places it breaks

Their report covers `validateEntityDirectory`. The same mismatch reaches two other call sites:

| Location (line numbers on `main`) | Symptom on Windows | Direction |
|---|---|---|
| `src/profile/paths.ts:75` | `template init` fails for every template — as reported | fail-closed |
| `src/profile/validate.ts:490` | any profile declaring a workflow `projectionFile` fails to load | fail-closed |
| `src/profile/paths.ts:58` | `pathsOverlap` fails to detect that a declared `wiki/` **contains** the reserved subtrees | **fail-open** |

The third is worth a maintainer's attention. On win32 the under-`wiki/` guard rejects every nested directory before `pathsOverlap` ever runs, so the only entity directory that loads at all is `wiki` itself: declared as `wiki/` or `wiki/.` it canonicalizes to `wiki`, clears the guard via the `child === dir` short-circuit, and then goes undetected as *containing* `wiki/graph`, `wiki/concepts`, and `wiki/outputs/workflows`. The one directory that is supposed to be forbidden is the only one that gets through. That is a validation bypass rather than a bad error message, which is why this PR treats it as more than a cosmetic fix.

## Why targeted rather than global

@squ1ddy proposed making `isInsideDir` separator-agnostic. I tested that patch: it cures all three sites, keeps every native confinement check correct on win32 including UNC paths, and passes 10 of 11 probes. The 11th is the reason I went narrower:

```
POSIX — '\' is a legal filename character on Linux/macOS
  current    isInsideDir("/p/sources\evil/secret.md", "/p/sources") = false   escape refused
  proposed                                                          = true    escape ALLOWED
```

Because that patch rewrites `\` to `/` before comparing, a directory literally named `sources\evil` beside `sources/` normalizes to look like it is *inside* `sources/`. `sources/` itself is not the reachable path — `confinedRegularFile` rejects symlinks at its `lstat` before `isInsideDir` runs. But several call sites do follow symlinks with `isInsideDir` as the sole gate, `src/export/okf/references.ts:57` among them, where an escaped file would be pulled into an OKF export. To be precise about blast radius: the escape target is a sibling inside the project root, so this is "reads a file the user never designated", not "reads `/etc/passwd`". Still a real bypass of the check whose stated purpose is preventing exactly that.

To be fair to the proposal, the regression comes from **rewriting inputs**, not from being global. A global fix that widened the separator set only on win32 — where `\` is not a legal filename byte, so both separators are unambiguous — would carry no POSIX risk. I went targeted anyway because it is four lines rather than a change to a helper that 27 security checks depend on. If you would rather have the general fix, I am happy to redo it that way.

The three broken sites do not need separator-agnosticism at all, though: their inputs are `/`-only by construction, so they just need to stop asking `path.sep` what the separator is. This PR adds `isInsidePosixDir` for those three and leaves `isInsideDir` untouched for the other 27 calls, all of which pass native absolute realpaths.

On POSIX `path.sep` is already `/`, so the new helper is the same function as before: no behaviour change on Linux/macOS.

## Tests

A grep gate, `test/profile-posix-path-gate.test.ts`, asserts the two lexical modules never reference `isInsideDir` or `path.sep` in code. It fails on `main` and passes here. Scoped to those two modules on purpose — `src/profile/scaffold.ts` and `src/profile/templates/publish/*` compare native realpaths and must keep using both.

`test/profile-posix-dir.test.ts` pins the property the argument above rests on: `isInsidePosixDir` does not normalize its inputs, so a literal backslash never matches. That property is unreachable through `validateEntityDirectory` — `normalizeDeclaredDir` strips backslashes first — so without a direct test, "hardening" the helper with a `\` → `/` rewrite would pass the entire suite *and* the gate while quietly reintroducing the escape.

Beyond those, the existing suites already cover the behaviour: `test/profile-workflows-validate.test.ts:139` asserts a `projectionFile` under `wiki/outputs/workflows/` is accepted, and `:156` asserts an entity directory overlapping that subtree is rejected. Both would fail on Windows today.

`test/genericity-grep-gate.test.ts` is the only pre-existing test this PR modifies. Both gates need the same comment-blanking helper — a gate that forbids a name has to ignore prose, or the docstring explaining *why* the name is forbidden trips it — and duplicating it broke the project's `fallow` duplication gate, so it moved to `test/fixtures/strip-comments.ts` and both import it.

## The reason this escaped

`.github/workflows/ci.yml` runs `ubuntu-latest` only, with a `node-version` matrix and no OS axis. At least 17 existing assertions across the two profile suites expect success and would already go red on win32 — starting with `baseProfile()` at `test/profile-validate.test.ts:37`. The tests were never the gap; the platform coverage was.

I have not added `windows-latest` here. That is a CI-spend decision, and since this codebase has never run on win32 the job may surface unrelated failures that shouldn't hold up this fix. Happy to open it as a follow-up PR if you want it.

## How to test

On Linux/macOS the fix is a no-op by construction — `path.sep` is already `/`, so `isInsidePosixDir` is literally the same function as the code it replaces. The suite is the regression net:

```bash
npm test                                    # 4344 passing
npx vitest run test/profile-posix-dir.test.ts test/profile-posix-path-gate.test.ts
```

To watch the gate genuinely fail rather than take my word for it, restore the two modules from `main` and re-run it:

```bash
git checkout main -- src/profile/paths.ts src/profile/validate.ts
npx vitest run test/profile-posix-path-gate.test.ts   # FAILS, reporting 5 code hits
git checkout HEAD -- src/profile/paths.ts src/profile/validate.ts
```

And the one that actually matters, on Windows — the check I could not run myself:

```powershell
mkdir probe; cd probe
llmwiki template init autosci
# on main:        Error: entity directory must be under 'wiki/': wiki/papers
# on this branch: succeeds
```

## What I could not verify

I have no Windows machine, and this deliberately does not mock `path` to simulate one. The reasoning and the POSIX-side behaviour are verified; **the actual win32 fix is not**. @squ1ddy — would you be willing to confirm this branch works on your setup? And if you already had a PR in flight, say so and I will close this in favour of yours with the two extra sites folded in.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,20 @@ on:
 env:
   FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 jobs:
+  # Linux only — a deliberate, revisitable choice, recorded here because it is
+  # what let issue #163 ship: on win32 `path.sep` ("\") made every nested profile
+  # directory fail containment, so no profile could load, and nothing here saw it.
+  #
+  # A windows-latest leg was tried in #172 and taken back out. It confirmed the
+  # separator fix holds on win32, then surfaced 64 unrelated failures across 44
+  # files — chiefly that O_NOFOLLOW is POSIX-only, so every "fails closed on a
+  # symlinked leaf" guard quietly degrades to a plain open there. Nor is
+  # `continue-on-error` a way to land the leg non-blocking: the workflow run goes
+  # green but the job's check-run still reports failure and blocks the merge.
+  #
+  # So win32 behaviour is held by the `sep` test seam in `src/utils/path-confine.ts`,
+  # NOT by CI. Restoring this leg means fixing those 64 failures first — worth
+  # doing as its own change, and not free.
   build-and-test:
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,22 +7,10 @@ on:
 env:
   FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 jobs:
-  # Runs on Linux AND Windows. llmwiki manipulates repo-relative paths
-  # everywhere, and a Linux-only matrix is what let issue #163 ship: on win32
-  # `path.sep` ("\") made every nested profile directory fail containment, so no
-  # profile could load, and nothing in CI could see it.
-  #
-  # The Windows leg is non-blocking (`continue-on-error`) for now: it reports the
-  # platform's real state without gating merges while the remaining win32 gaps
-  # are worked through. Flip it to required — delete the `continue-on-error`
-  # line — as soon as it is green, or it decays into an ignored red check.
   build-and-test:
-    runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.os == 'windows-latest' }}
+    runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
         node-version: [24]
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,10 +7,22 @@ on:
 env:
   FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 jobs:
+  # Runs on Linux AND Windows. llmwiki manipulates repo-relative paths
+  # everywhere, and a Linux-only matrix is what let issue #163 ship: on win32
+  # `path.sep` ("\") made every nested profile directory fail containment, so no
+  # profile could load, and nothing in CI could see it.
+  #
+  # The Windows leg is non-blocking (`continue-on-error`) for now: it reports the
+  # platform's real state without gating merges while the remaining win32 gaps
+  # are worked through. Flip it to required — delete the `continue-on-error`
+  # line — as soon as it is green, or it decays into an ignored red check.
   build-and-test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.os == 'windows-latest' }}
     strategy:
+      fail-fast: false
       matrix:
+        os: [ubuntu-latest, windows-latest]
         node-version: [24]
     steps:
       - uses: actions/checkout@v4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Windows: broken links in the generated wiki index** — the same separator bug on the output side. Entity-page links in `wiki/index.md` are built from `path.relative`, which emits `\` on win32, so a NESTED entity directory produced the unusable link `research\papers/foo.md`. Link targets are now normalized to POSIX. Single-level directories were unaffected, which is why this went unnoticed (#163).
 
-### Changed
-
-- **CI now runs on Windows as well as Linux.** A Linux-only matrix is what allowed the two path-separator bugs above to ship. The Windows leg is non-blocking for now so it reports the platform's real state without gating merges, and should be made required once green.
+- **`npm test` could not run on Windows at all** — vitest's global setup shelled out to `npx`, which is `npx.cmd` there and has not been resolvable by `child_process` without `shell: true` since the Node 22 hardening for CVE-2024-27980. The setup threw, collection aborted, and vitest reported the unrelated "No test files found". It now invokes the build directly with the running Node binary, needing no shell on any platform.
 
 ## [1.1.0] - 2026-07-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Windows: profile path validation rejected every declared directory** — on win32, `llmwiki template init` failed for every template with `entity directory must be under 'wiki/'`, any profile declaring a workflow `projectionFile` failed to load, and an entity directory declared as `wiki/` was wrongly accepted despite containing every reserved subtree — on win32 it was the only entity directory that loaded at all. Declared directories canonicalize to `/`-joined repo-relative paths, but the containment check built its prefix with the platform separator (`\` on Windows), so no nested path ever matched. The lexical profile-path checks now compare POSIX paths directly; native path confinement is unchanged. Reported and diagnosed by @squ1ddy (#163).
 
+- **Windows: broken links in the generated wiki index** — the same separator bug on the output side. Entity-page links in `wiki/index.md` are built from `path.relative`, which emits `\` on win32, so a NESTED entity directory produced the unusable link `research\papers/foo.md`. Link targets are now normalized to POSIX. Single-level directories were unaffected, which is why this went unnoticed (#163).
+
+### Changed
+
+- **CI now runs on Windows as well as Linux.** A Linux-only matrix is what allowed the two path-separator bugs above to ship. The Windows leg is non-blocking for now so it reports the platform's real state without gating merges, and should be made required once green.
+
 ## [1.1.0] - 2026-07-15
 
 Adds a security-first template distribution ecosystem — publishers sign and distribute profile templates offline, and consumers discover, install, update, and verify them through explicitly trusted taps — plus a guided path for authoring a first profile and the `llmwiki status` command. Every addition is opt-in; projects that do not use templates, taps, or profiles are unaffected.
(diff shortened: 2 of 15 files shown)


ethanj (changes requested):
Thanks for this — the diagnosis is the good part. Pinning the failure on `"wiki/papers".startsWith("wiki\\")` and then resisting the obvious fix is exactly right: making `isInsideDir` separator-agnostic would have traded a Windows bug for a POSIX escape on the 27 call sites that pass native realpaths, since a backslash is a legal filename character there. Adding a narrow `isInsidePosixDir` for the three declared-path call sites and leaving the confinement helper alone is the correct shape, and I verified it holds — the new helper never normalizes its input, so a literal backslash in a POSIX name still fails to match, which fails closed rather than open. The grep gate to keep the two from being confused again is a good instinct.

One same-class defect is still live, plus two smaller things.

1. Requesting changes — native separators still reach public paths

The fix covers containment comparisons, but not every place a path becomes user-visible output. `toEntityProblemView` returns `path.relative` raw, and its own JSDoc calls the result "the public, path-safe problem view":

https://github.com/atomicstrata/llm-wiki-compiler/blob/3e6b01e0ab1160ac9c2b93e4b1b11346fdf70393/src/profile/types.ts#L462-L470

`relativizeFile` has the same shape:

https://github.com/atomicstrata/llm-wiki-compiler/blob/3e6b01e0ab1160ac9c2b93e4b1b11346fdf70393/src/profile/artifact-lint.ts#L236-L239

On win32 `path.relative` returns native separators by definition, so status and viewer output become `wiki\notes\untitled.md` where the contract promises `wiki/notes/untitled.md` — the same class of bug as #163, one layer out. `toPosixPath` is the fix you already wrote; these just need to route through it.

Worth noting why this survived: the new gate asserts only that those modules never reference `isInsideDir` or `path.sep`, so raw `path.relative` output passes straight through it.

https://github.com/atomicstrata/llm-wiki-compiler/blob/3e6b01e0ab1160ac9c2b93e4b1b11346fdf70393/test/profile-posix-path-gate.test.ts#L66-L72

A gate that catches the mechanism rather than the two named symbols would cover this and whatever the next one is.

2. Minor — stale comment about a CI job this PR removes

`toPosixPath`'s JSDoc justifies the `sep` parameter as a test seam that keeps win32 behavior "verified until the Windows CI job is green", but the final commit drops the `windows-latest` leg, so there's no such job left. You updated the two sibling mentions of this (`test/global-setup.ts` and the CHANGELOG) and this one got missed.

https://github.com/atomicstrata/llm-wiki-compiler/blob/3e6b01e0ab1160ac9c2b93e4b1b11346fdf70393/src/utils/path-confine.ts#L52-L60

3. Not a change request, just flagging it explicitly

The Windows CI leg is added in one commit and removed in another, so `gh pr diff` shows zero net change to `ci.yml` and a reviewer reading only the diff would never know the question was raised. Your reasoning for dropping it is sound — 64 unrelated failures, and `continue-on-error` not behaving as expected — but the net effect is that a Windows-only fix ships with no Windows coverage, which is the condition the earlier commit correctly identified as why #163 shipped in the first place. I'd rather that be a decision on the record than a silent one. Happy for it to stay out of this PR; I just don't want it to disappear.

Item 1 is the one I'd like fixed before this lands. 2 is quick, and 3 is yours to call.

PipDscvr:
Both change requests are fixed, along with the gate change behind item 1. Pushed as 62ed923 and 8ec204b.

## 1. Native separators reaching public paths — fixed

Both producers now route through `toPosixPath`:

- `toEntityProblemView` — `src/profile/types.ts`
- `relativizeFile` — `src/profile/artifact-lint.ts`

I checked whether the two you named were the whole set rather than assuming. Inside `src/profile/**` they are: the only other `path.relative` call sites are in `templates/publish/`, already on the gate's `NATIVE_MODULES` list and correctly building native paths. Outside the profile layer there are two more — `src/utils/atomic-write.ts:74` and `src/trust/journal.ts:396` — and both feed their result straight back into `confineUnderRoot` as a filesystem argument, so native is correct there and I left them alone.

**The gate now catches the mechanism.** Your diagnosis of why this survived is the actionable part: naming `isInsideDir` and `path.sep` only ever catches the two symbols the first fix happened to touch. The third rule gates the operation instead — `path.relative` is the call that turns a native filesystem path into portable content, so in this layer its result must always be routed through `toPosixPath`.

Reverting just the two fixes turns it red and names both lines:

```
- Expected
+ Received

- []
+ [
+   "artifact-lint.ts:238 — return path.isAbsolute(file) ? path.relative(canonicalRoot, file) : file;",
+   "types.ts:467 — ...(problem.filePath !== undefined ? { path: path.relative(root, problem.filePath) } : {}),",
+ ]
```

`path.join` is deliberately **not** gated, and the reason is in the file so the next person doesn't "tighten" it: its ~23 call sites in this layer build native paths to hand back to `fs`, which is exactly what it should emit. The invariant is about values that leave as content, not every touch of the platform separator. A matcher self-test pins the blanking logic — including a line carrying both a wrapped and a bare call — since a gate whose regex is quietly wrong is worse than no gate at all.

## 2. Stale comment — fixed

`toPosixPath`'s docstring no longer justifies its `sep` parameter against a job that doesn't exist. The seam isn't a stopgap until Windows CI goes green; it's the only win32 coverage there is, which is what it now says.

## 3. Windows CI — on the record, still out of the PR

Taking your framing: the call stands, but it shouldn't be invisible. It's now a comment in `ci.yml`, so the diff carries it instead of a commit message the squash would bury — why Linux-only is what let #163 ship, what the `windows-latest` leg proved and what it surfaced (64 unrelated failures across 44 files, chiefly `O_NOFOLLOW` being POSIX-only, so every "fails closed on a symlinked leaf" guard silently degrades to a plain open there), why `continue-on-error` can't land it non-blocking, and that restoring it means fixing those 64 first. No job added — this is the decision written down, not reversed.

## Verification

`npx tsc --noEmit`, `npm run build` and `fallow` all clean. Profile and path suites green — 91 tests across `profile-posix-path-gate`, `profile-posix-dir`, `posix-path`, `profile-workflows-validate`, `profile-validate` and `genericity-grep-gate`.

Full suite: 6 failures, all pre-existing. I confirmed they reproduce identically with my changes stashed — credential-dependent tests that pass only with no provider configured, plus a flaky `workflow-active-run-quota` trio. None are in this area, and none are affected by this change.

Unchanged from the original post: I still have no Windows machine. The POSIX behaviour and the reasoning are verified; the win32 behaviour is not.

ethanj (approved):
Approving — this is the right response to the review.

The part I want to call out is the third gate rule. I flagged two call sites; you fixed those and then went after the thing that let them through, so the gate now catches any `path.relative` result leaving this layer unwrapped rather than the two symbols the original fix happened to touch. The matcher self-test that pins the both-forms-on-one-line case is what makes it a control rather than a comment, and deliberately leaving `path.join` ungated — with the reason written down — is the judgement that keeps it from becoming noise people route around.

Recording the Linux-only decision in `ci.yml` is the other thing worth having. The 64 failures and the `continue-on-error` behaviour are exactly the kind of finding that evaporates if it only lives in a commit message, and the next person to reach for a Windows leg now starts where you left off instead of rediscovering it.

PR #177, opened 2026-08-07
feat(viewer): Nebula redesign, profile-aware navigation, and the data it needed

### LINKED ISSUE #176: Nebula viewer redesign: profile-aware navigation, and the data three of the screens need
Opening this ahead of a large PR so the scope can be argued with before anyone reads 125 files.

## What

Implement the **Nebula** design direction for `llmwiki view`, from four Claude Design files: the overall direction, the design system, the health screen, and a two-part proposal for profile vocabulary and pipeline.

Working rule throughout: **no invented data.** Where a mockup asked for something the codebase couldn't supply, it was flagged rather than faked. Twice that meant building the data first; several elements were deliberately not built at all.

## Why it grew past a restyle

Three of the four designs turned out to need data that did not exist or did not reach the client:

- The health screen's Lint panel wanted a **per-rule breakdown**. `writeLintCache` persisted only two integers, so the rule, file and message never left the CLI process.
- The profile-vocabulary sidebar wanted to project a profile's **entity types**. Typed entity pages were collected at snapshot time but fed only to the graph — `/api/pages` returned 0 pages for a newsroom project with articles on disk.
- The Pipeline panel wanted **lifecycle states, transitions and relation endpoints**. All of it existed on the profile pack; none of it reached the viewer.

So the change is roughly half redesign and half the data plumbing those screens require.

## Pre-existing bugs found while building

These weren't the goal, and I'd rather flag them here than have them read as unrelated scope in the PR:

- **`/api/health` leaked absolute paths.** Lint aggregates carried on-disk paths, and that route doesn't participate in the non-loopback suppression `/api/page` and `/api/index` use — so `llmwiki view --allow-lan` served the user's home directory to the network. Fixed at the source rather than by gating the endpoint.
- **Exports were non-deterministic.** `export/collect.ts` invented `new Date()` for any page lacking `updatedAt` — which is every saved query — propagating into six formats. An export → import round trip then baked the fabricated timestamp into frontmatter on disk. A golden test was masking it by normalising volatile dates.
- **Saved queries rendered wrong everywhere.** `query --save` writes `type` + `createdAt`; the viewer read `kind` + `updatedAt`. Every saved query showed a blank age, was labelled a concept, and sorted last in "recently compiled". Fixed on the reader side so existing wikis heal without recompiling.

Two consumers still share that last root cause and are **reported, not fixed**: `src/context/graph.ts` buckets saved queries as concepts for agent-context edge scoring, and `src/linter/rules-crosslinks.ts` would emit a false `schema-cross-link-minimum` under a non-`concept` `defaultKind`.

## Deliberately not built

Each because the data or the surface does not exist:

- `Settings` and `Compile & export` nav entries — no such surface in a read-only viewer
- Every mutation affordance the mockups drew (per-row fix verbs, "Resolve dangling links", gate approvals) — the viewer is a frozen read-only snapshot, so these became navigation, or name the CLI command as text
- `--require-citations` (not a flag we ship), `148 lines` (not collected), `Claims` (no claims inventory — citations substituted)
- Time-in-state on the Pipeline panel — no per-page state timestamp exists anywhere

## One thing worth a maintainer's opinion

The profile-vocabulary design makes BROWSE a projection of the active profile, so nav labels become per-project: a newsroom project reads *Articles / Desks / Bylines* where a default project reads *Concepts / Queries*. The designer names the cost explicitly — docs and shortcuts can no longer say "click Concepts".

Default projects are unchanged byte-for-byte, and that's pinned by goldens captured from the pre-change build. But it is a real change to how the product is described, and it seemed worth raising before it lands rather than after.

## Testing

`4836 passed, 0 failures` against current `upstream/main` (merged in) · `tsc --noEmit` clean · `npm run fallow:ci` clean.

The three security suites — `viewer-host`, `viewer-render-security`, `viewer-path-safety` — pass **unmodified** throughout; `viewer-path-safety` only gained cases. Every UI change was verified in Chrome against a real project in both themes, not only under JSDOM.

## Videos

<video src="https://github.com/user-attachments/assets/e5ac3606-dec4-4bf0-b8e4-01281aa4d23a"></video>

<video src="https://github.com/user-attachments/assets/2b3351f0-3979-49c5-b943-ef32677311a0"></video>

PR to follow.

PipDscvr:
Closes #176.

Implements the **Nebula** design direction for the local viewer, from four Claude Design files, plus the data work several of those screens turned out to need.

The brief was "no new features unless the data already exists". Where a mockup asked for something the codebase couldn't supply, it was flagged rather than faked — twice that meant building the data first (per-rule lint aggregates; typed entity pages reaching the viewer), and several elements were deliberately not built at all.

## What's in it

**The redesign** — sidebar, header, dashboard, content and graph restyled to the Nebula system; self-hosted Space Grotesk / JetBrains Mono; light and dark themes from one token set; full-bleed layout with a sticky sidebar and header.

**New surfaces**
- `#/concepts`, `#/queries`, `#/sources` list routes
- `#/health` — rebuilt as the Nebula health screen: a divided CONTENTS strip, a Lint panel with a per-rule breakdown, Freshness, Traceability
- `#/reviews` — the pending review queue (the sidebar entry previously pointed at `#/health`)
- `#/workflows` — parked workflow runs, each naming the CLI command that unblocks it
- `#/pipeline` — a profile's lifecycle states, ordered by the declared transition graph

**Profile (CLP) support** — typed entity pages now reach the viewer's page surfaces, are searchable and routable; BROWSE projects the profile's declared entity types; profile collector problems surface in the header verdict and on `#/health`.

**Data work the screens required**
- `llmwiki lint` persists a bounded per-rule breakdown (`rules[]`) in its cache
- `/api/pages` carries the profile's entity counts, lifecycle definitions and relation types
- `/api/reviews` and `/api/workflow-runs` are consumed (the latter existed with no client)

## Fixes found along the way

Several were pre-existing, found while building:

- **`/api/health` leaked absolute paths.** Lint rule aggregates carried on-disk paths, and that endpoint doesn't participate in the non-loopback suppression `/api/page` uses — so `view --allow-lan` served the user's home directory. Fixed at the source.
- **Exports were non-deterministic.** `export/collect.ts` invented `new Date()` for pages lacking `updatedAt` (every saved query), across six formats; an OKF round trip baked the fabricated time into frontmatter. A golden test was masking it.
- **Saved queries rendered wrong everywhere.** `query --save` writes `type` + `createdAt`; the viewer read `kind` + `updatedAt`. Every saved query showed a blank age, was labelled a concept, and sorted last. Fixed on the reader side so existing wikis heal without recompiling.
- **Graph nodes 404'd on click** for typed pages — the click built its hash from an on-disk directory whose separator encoded to `%2F`.
- **The header asserted freshness it never measured** — with no `state.json`, nothing can be stale, so "ALL PAGES FRESH" appeared over a project whose freshness was unverifiable. The pill now reports the whole-wiki verdict and has a third state for "unknown".

## Deliberately not built

Flagged rather than implemented, each because the data or the surface doesn't exist:

- `Settings` and `Compile & export` nav entries — no such surface in a read-only viewer (the designer has since dropped them from the mockups)
- Mutation affordances — "Resolve dangling links", per-row fix verbs, gate approvals. The viewer is a frozen read-only snapshot; these became navigation, or name the CLI command as text
- `--require-citations`, `148 lines`, `Claims` — a flag that doesn't exist, data that isn't collected, and a claims inventory the compiler doesn't produce (citations substituted, per spec §5.3)
- Time-in-state on `#/pipeline` — no per-page state timestamp exists anywhere

## Testing

`4836 passed, 0 failures` (against current `main`, merged in) · `tsc --noEmit` clean · `npm run fallow:ci` clean — the diff-scoped check CI runs.

The three security suites — `viewer-host`, `viewer-render-security`, `viewer-path-safety` — pass **unmodified** throughout; `viewer-path-safety` only gained cases.

Every UI change was verified in Chrome against a real project, in both themes, not only under JSDOM. Default-profile output is pinned byte-identical (envelope, page routes, search, graph and sidebar) by goldens captured from the pre-change build.

**Note for reviewers running the suite:** `env -u LLMWIKI_PROVIDER -u LLMWIKI_MODEL npx vitest run`. Six credential tests assert "no provider configured" by reading ambient env, so they fail in any shell that exports `LLMWIKI_PROVIDER`. Also avoid two concurrent full runs — vitest's `globalSetup` runs `tsup`, which cleans `dist/` mid-flight.

## Known limitations

- `EntityTypeDef.titleField` is in the schema but consumed nowhere, so typed pages display their slug rather than a title
- Typed pages carry no freshness (no `sources` + `state.json` participation), so "Recompile N stale pages" can never fire on a profile project
- `src/context/graph.ts` and `src/linter/rules-crosslinks.ts` share the saved-query kind bug fixed here in the viewer and export paths — reported, not fixed
- The nav labels a profile projects are per-project, so docs can no longer say "click Concepts" — an accepted cost of the chosen design

## Not included

`examples/` is byte-identical to `main`. Two example projects were built for verification and removed at the author's request.


ethanj (changes requested):
This is strong work, and the discipline behind it shows in places that are hard to fake. Opening #176 first so the scope could be argued before anyone read 125 files was the right call and made this reviewable. The "no invented data" rule is visible throughout — the Deliberately not built list is the most useful section of the PR, and flagging `--require-citations`, `148 lines`, and Claims as things the codebase cannot supply is exactly the judgement I want from a contributor handed a mockup.

I checked the claims rather than taking them: the three security suites really do pass unmodified (the one removed line in `viewer-path-safety` is the `vitest` import, replaced by one adding `afterEach` — no assertion weakened), 4836 pass with 3 skipped on my own run, type-check and diff-scoped fallow are clean across all 125 files. The path handling got genuinely stronger — `/api/page`'s directory allowlist is now derived from the profile's declared entity types instead of two hardcoded literals, and the handler never touches the filesystem at all. `/api/reviews` slices the id list before reading any file, so a 5,000-candidate queue still costs 200 reads. New DOM goes through `textContent` helpers throughout.

So the problems below are not craft problems. They are all one shape, and it is worth naming because the fix is a single pattern applied in a few places.

## The pattern

**A server-side projection drops a field that exists, the client substitutes a default, and the default is presented to the user as fact.** That is the inverse of the rule this PR set for itself: not invented data on a mockup, but invented data at runtime. Three of the four below are instances; the fourth is the same drift running the other way.

## 1. Blocking — workflow rows recommend commands that cannot unblock the run

`toRunRow` projects eight fields and drops `awaitingTrustGate`, `nextSubmitEntityType`, and `nextSubmitArtifactType`:

https://github.com/atomicstrata/llm-wiki-compiler/blob/43b1f49b733351a68165afd41457847b0b8fbe7b/src/viewer/workflow-runs.ts#L42-L54

Those three are not incidental. Read what `RunStatus` says they are for:

https://github.com/atomicstrata/llm-wiki-compiler/blob/43b1f49b733351a68165afd41457847b0b8fbe7b/src/workflows/status.ts#L64-L97

`awaitingTrustGate` exists, in its own words, to "let a renderer hint the trusted-write grant + re-submit instead of a `gate approve` that would fail." `nextSubmitEntityType` and `nextSubmitArtifactType` exist to supply "a concrete, valid `--entity-type` value an operator/agent can submit against." They were added for precisely this renderer, and this renderer is the one consumer that does not receive them.

The consequences are both live:

https://github.com/atomicstrata/llm-wiki-compiler/blob/43b1f49b733351a68165afd41457847b0b8fbe7b/src/viewer/assets/viewer-workflows.js#L195-L204

- `llmwiki workflow submit <run-id>` is printed with no `--kind`, and `buildStageOutput` calls `requireOption(options.kind, "--kind")` before anything else, so the command fails immediately on paste.
- `llmwiki workflow gate approve <run-id> <gate>` is printed whenever `awaitingGate` is set, including trust gates — and `vouchGate` throws `TrustGateNotHereError` for exactly that case.

A parked run is the moment a user most needs a correct instruction, and the row currently guarantees a failed one. Carrying the three fields through the projection and branching on them is the whole fix.

## 2. Should fix — typed review candidates are shown as concepts

`toReviewRow` copies `targetDirectory` and drops `targetEntityType`:

https://github.com/atomicstrata/llm-wiki-compiler/blob/43b1f49b733351a68165afd41457847b0b8fbe7b/src/viewer/reviews.ts#L86-L100

The renderer then falls back to a literal `"concepts"` for anything without a directory:

https://github.com/atomicstrata/llm-wiki-compiler/blob/43b1f49b733351a68165afd41457847b0b8fbe7b/src/viewer/assets/viewer-reviews.js#L145-L158

`review-approve.ts` routes a typed candidate by `targetEntityType`, so for those candidates the viewer states a destination that is not where approval writes. This is the human review gate — the one screen whose job is to tell a reviewer what they are about to accept.

## 3. Should fix — the typed-list empty state sends authors to a directory the collector may not scan

`entityTypeDefinition` projects `type` and `lifecycle`, omitting the entity type's declared `directory`:

https://github.com/atomicstrata/llm-wiki-compiler/blob/43b1f49b733351a68165afd41457847b0b8fbe7b/src/viewer/pipeline.ts#L103-L118

The client reconstructs it from the type id, and its comment asserts the identity outright — "it is also the directory name the reader has to create":

https://github.com/atomicstrata/llm-wiki-compiler/blob/43b1f49b733351a68165afd41457847b0b8fbe7b/src/viewer/assets/viewer-lists.js#L87-L100

`EntityTypeDef.directory` is a required, independently declared field, and `profile/collect.ts` scans `def.directory`, not the type id. So for any profile where the two differ — `ideas` stored under `wiki/ideas-v2/`, say — the empty state instructs an author to create pages in a directory the collector never reads. They author content, nothing appears, and the screen that misdirected them is the same one still showing empty. Wrong guidance that produces invisible work is worse than no guidance.

## 4. Should fix — absent timestamps become present-but-empty metadata in five export formats

The determinism fix is right, and killing the fabricated `new Date()` was worth doing. But the sentinel it leaves behind is rendered unconditionally:

https://github.com/atomicstrata/llm-wiki-compiler/blob/43b1f49b733351a68165afd41457847b0b8fbe7b/src/export/collect.ts#L129-L133

https://github.com/atomicstrata/llm-wiki-compiler/blob/43b1f49b733351a68165afd41457847b0b8fbe7b/src/export/json-ld.ts#L22-L31

https://github.com/atomicstrata/llm-wiki-compiler/blob/43b1f49b733351a68165afd41457847b0b8fbe7b/src/export/llms-txt.ts#L31-L39

OKF omits the field correctly, but JSON-LD emits `"dateCreated": ""`, and llms.txt emits `created:  | updated:`. An undated page now asserts an empty date rather than declining to state one — which for JSON-LD in particular is a schema-invalid value a consumer must special-case. Making the fields optional and having each renderer omit an absent value finishes the fix you started.

## Minor

**A payload nothing reads, described by comments saying it does.** `server.ts` embeds a `#page-index` JSON blob into every served HTML page, carrying id, directory, slug, title, kind and entityType for every page in the wiki. On `main` the consumer is `viewer.js`'s `PAGE_INDEX_SELECTOR`; the sidebar rewrite deleted it, and no client file reads the element at this head. `shell.ts` still states the blob is "used by the sidebar to group concepts on first paint" and carries "the fields the client needs for first-paint sidebar rendering."

https://github.com/atomicstrata/llm-wiki-compiler/blob/43b1f49b733351a68165afd41457847b0b8fbe7b/src/viewer/shell.ts#L28-L38

Worth noting the commit that surfaces typed entity pages edits this comment block and adds `entityType` to the payload — extending something already unreachable, after the consumer was gone.

**The router docblock in `viewer.js` omits `#/graph`, `#/concepts`, `#/queries` and `#/sources`**, all four of which are in the same file's route map.

**Two docs the redesign invalidated but did not touch**: `docs/cli/lint-eval.mdx` still says the viewer shows health counts in the sidebar, and `docs/troubleshooting/stale-pages.mdx` still describes a sidebar health pane and a sidebar freshness filter. Both moved to `#/health` and `#/concepts`. `git diff` on those files is empty, so no changed-files check would catch them — the redesign is simply what made them wrong. `docs/cli/view.mdx` itself is thorough.

## On your maintainer question

Making BROWSE a projection of the active profile is the right call and I would not trade it back for stable nav labels. Default projects are pinned byte-identical, the cost is confined to prose about profile projects, and "click the entity type your profile declares" is a fair thing for docs to say. Go ahead.

Item 1 is what I would like fixed before this lands; 2 through 4 are small and sit on the same seam, so they are probably one sitting. The minor items are yours to take or leave.

PipDscvr:
All four are fixed in `fe3b779`, plus all three minor items. Naming the pattern was the useful part of the review — once "the projection drops a field, the client invents a default" was on the table, three of the four fixes were the same edit in three places, and the fourth turned out to be the same failure one layer later. Details per item.

---

## 1. Blocking — workflow rows

Carried all three fields through, and branched on them.

`toRunRow` now delegates to a `copyParkHints` helper that mirrors `applyParkHints` in `status.ts` field-for-field, so the two cannot drift:

```ts
function copyParkHints(row: WorkflowRunRow, status: RunStatus): void {
  if (status.awaitingGate !== undefined) row.awaitingGate = status.awaitingGate;
  if (status.awaitingTrustGate === true) row.awaitingTrustGate = true;
  if (status.awaitingOutput !== true) return;
  row.awaitingOutput = true;
  if (status.nextSubmitEntityType !== undefined) row.nextSubmitEntityType = status.nextSubmitEntityType;
  if (status.nextSubmitArtifactType !== undefined) row.nextSubmitArtifactType = status.nextSubmitArtifactType;
}
```

On the client, `nextCommands` and a new `submitCommand` mirror `workflow-shared.ts`'s `nextHintOf`/`submitCommand`, so the viewer and `workflow status` now print the *same* command for the same parked run:

- **Output park** → `$ llmwiki workflow submit run-0002 --kind page --entity-type articles --slug <slug> --body-file <path>`
- **Artifact-only stage** (declares `artifactWrites`, no `writes`) → `--kind artifact --artifact-type result …`
- **Trust gate** → no `gate approve` line at all. The row gets a `.workflow-note` naming what actually clears it — `LLMWIKI_TRUSTED_WRITE` plus a re-submit — and the submit command below it. The chip also reads `Trust gate · <id>` rather than `Awaiting gate · <id>`; calling both "awaiting gate" was the thing pointing the reader at the call that throws.

A stage that declares writes always supplies `nextSubmitEntityType`, so the `<entity-type>` placeholder is only reachable on a malformed row — it is kept as the defensive default because a visibly-unfilled placeholder beats a silently-missing flag.

**Tests.** New `test/viewer-workflow-run-hints.test.ts` pins the three fields onto the wire row, including the exact key set (the row must stay status-only — no `run`, no `storeLevel`). `viewer-workflows-route.test.ts` gained the artifact-only and trust-gate fixtures and asserts the full command strings; the old assertion `expect(next?.textContent).toBe("$ llmwiki workflow submit run-0002")` was pinning the broken output, so it is replaced rather than added to.

## 2. Review candidates

`ReviewCandidateRow` carries `targetEntityType`, and the renderer reads it *first* — matching `routeApprovedPageWrite`, which branches on it before anything else:

```js
function targetDirectoryOf(review) {
  if (typeof review.targetEntityType === "string" && review.targetEntityType.length > 0) {
    return review.targetEntityType;
  }
  return review.targetDirectory || DEFAULT_TARGET_DIRECTORY;
}
```

A typed candidate now reads `crossref.json → wiki/papers/`. Tests cover all three branches — typed, `targetDirectory`, and neither — on both legs (`viewer-reviews-api.test.ts` for the projection, `viewer-reviews-route.test.ts` for the render).

## 3. Typed-list empty state

`PipelineEntityTypeDef` carries `directory`. The lookup moved to `viewer-nav-types.js` as `typeDirectory(entityTypes, type)` — it sits beside `navTypeLabel` because both answer "how does this declared type appear to a reader", and both have to read the profile rather than guess from the id.

The empty state now says `Author them as Markdown under wiki/ideas-v2/` for a profile that declares `ideas` with `directory: "ideas-v2"`. When the envelope carries no declaration it says "in the directory your profile declares" and names no path — you were right that wrong guidance is worse than none, and that applies to my fallback too.

`viewer-pipeline-envelope.test.ts` asserts every row's `directory` equals what the profile declared (not a hardcoded list), plus a divergent case through `buildPipelineDefinitions` directly. `viewer-entity-list-route.test.ts` covers the render for divergent, matching, and absent.

## 4. Export timestamps

`ExportPage.createdAt`/`updatedAt` are optional; `readPageTimestamps` omits the keys rather than degrading to `""`. You framed this as finishing the fix I started, which is right — emitting `""` only moved the invention one layer downstream, where "the date is the empty string" is a claim the page never made.

Five formats changed:

| Format | Before | After |
|---|---|---|
| JSON | `"createdAt": ""` | key absent |
| JSON-LD | `"dateCreated": ""` | property absent |
| GraphML | `<data key="createdAt"></data>` | element absent |
| llms.txt / llms-full.txt | `created:  \| updated:` | clause absent |
| Marp | `<!-- created:  \| updated: -->` | clause absent; comment dropped if nothing remains |

OKF was already correct and is unchanged. The shared decision lives in one place (`src/export/timestamps.ts`) so llms.txt and Marp cannot drift on it; the two structural formats omit a key, which is their own idiom.

`export-timestamp-fidelity.test.ts`'s `AFFECTED_FORMATS` table gained a `createdMarker`, so passing `""` to a marker yields exactly how that format *would* spell an empty timestamp — and a new test asserts none of those strings appear for an undated page. The undated page renders alone there, because the dated fixtures carry the same substrings as prefixes of their real values. `docs/cli/export.mdx` documents the behaviour and the two new per-page fields.

## Minor items — all three taken

**The `#page-index` blob is gone**, not re-commented. Confirmed against `main` first: `PAGE_INDEX_SELECTOR` was live there and has no reader at this head. Keeping it meant serializing the whole page list into every HTML response for nothing, and a second copy of the page list on the wire is a second thing that can disagree with `/api/pages`. Removed the marker from `index.html`, `substitutePageIndex` from `shell.ts`, the call from `handleShell` (which no longer needs the snapshot), and the fixture's `embedPageIndex` + its `pages` parameter. `shell.ts` now explains what used to be there and why it is not — a reader who wonders where first-paint data went should not have to check git.

You were also right that the typed-pages commit extended it. That is fixed by the removal rather than by trimming `entityType` back out.

The test fixture's `EmbeddedPage` is renamed `PageRow`: those suites still need the shape, but it is an `/api/pages` row now, not a shell row, and `viewer-rail.test.ts`'s `embedded` / `STICKY_EMBEDDED` locals were renamed to match. `viewer-shell-template.test.ts` lost the escape-contract tests along with the thing they protected, and gained one asserting the template is returned byte-for-byte with nothing substituted.

**The router docblock** now lists all nine static routes and points at `STATIC_ROUTES` as the map it must agree with. It also notes that the four list routes and the graph read the bootstrap envelope and issue no request of their own, since that is the fact the old text elided.

**Both stale docs are fixed.** `lint-eval.mdx` now says the sidebar's Health & lint entry carries the finding count as a badge and the `#/health` screen shows the totals with the per-rule breakdown. `stale-pages.mdx` now says the badges are on an open page's metadata rail, that `#/concepts` carries the per-row freshness dot and the filter, and that `#/health` has the aggregates. I checked each claim against the code rather than editing around the wrong words — the corrupt-state banner does still exist, so that sentence stands, with the header's unverified verdict added.

---

## Verification

`npx tsc --noEmit` clean · `npm run build` clean · `4859 passed, 3 skipped, 0 failures` · `npm run fallow:ci` clean.

The three security suites (`viewer-host`, `viewer-render-security`, `viewer-path-safety`) are byte-identical to what you reviewed — `git diff 43b1f49` on all three is empty.

Two notes on the shape of the work:

- Fallow flagged `nextCommands` and the directory lookup at cyclomatic 6 on the first pass. Both were split into named predicates (`needsSubmit`, `needsGateApproval`, `typeDirectory`) rather than suppressed — they read better as predicates anyway.
- These are data-projection fixes and I verified them at the DOM level rather than in Chrome. Reproducing a parked trust-gated run in a live project is a fair amount of setup for a code path the JSDOM tests pin exactly; flagging it rather than implying a browser pass I did not do.

Not touched: #178. Nothing here adds a `profileSchema`, a relation drilldown, an artifact route, or a field-format vocabulary — `PipelineEntityTypeDef.directory` is the entity type's own declared directory, which the empty state needed, and it is not the start of that read model.

PipDscvr:
Correction on item 3, pushed as `599e283`.

My fix rendered the declared directory as `wiki/${directory}/`. `EntityTypeDef.directory` is already a canonical **project-relative** path — every shipped template spells it `wiki/<name>`, and `scanEntityDir(root, def.directory)` resolves it against the project root, not against `wiki/` — so AutoSci's `papers` came out as `wiki/wiki/papers/`.

Same failure mode as the one you reported, one layer over: an author sent to a directory the collector never scans. The renderer now prints the declared path verbatim.

It survived my tests because the vocabulary fixture's `types()` helper defaulted `directory` to a bare type id, which no real profile declares. That default is now `wiki/<type>`, and there is a case pinning a directory outside `wiki/` so nothing is prepended there either — a synthetic bare name can no longer hide a prefix bug.

`tsc` / build / `4860 passed, 3 skipped, 0 failures` / `fallow:ci` all clean on the new head. Nothing else in the previous push changed.

ethanj (approved):
Approving — and the response went after the seam rather than the four symptoms.

Making `directory` required on `PipelineEntityTypeDef` is the best of these: that projection can no longer lose the field without failing to compile, so the class is closed structurally rather than pinned by a test someone can delete. Mirroring the CLI's `submitCommand` semantics instead of inventing a second set means the two surfaces can't print different instructions for the same parked run, and carrying the trusted-write grant into the trust-gate note is the part I would have asked for next — a re-submit hint without it just re-parks the run and tells the operator nothing about why.

Two of the fixes are better than the findings that prompted them. Routing the review target on `targetEntityType` first doesn't just display something truer, it matches what approval actually does. And making the export timestamps optional at the type level is a real fix where patching five renderers would have been a patch — the shared clause helper means llms.txt and Marp can't drift apart on the same question later.

The `#page-index` removal is worth calling out too. Deleting the payload was the ask; keeping a past-tense note explaining why it existed and why it wasn't retained "in case" is what stops someone re-adding it in six months, and the reason given — a second copy of the page list being a second thing that can disagree with `/api/pages` — is the right one.

Good PR. Opening #176 to argue the scope before writing any of it is what made 125 files reviewable at all, and the Deliberately not built list did more for my confidence than any amount of green CI would have.

PR #179, opened 2026-08-12
feat(cli): add llmwiki rm <source>

### LINKED ISSUE #60: Add `llmwiki rm <source>` to delete a source and its derived concepts
Cleaning up a bad ingest (e.g. URL returned an error page) currently takes 3 manual steps:

1. `rm sources/<bad>.md`
2. `llmwiki compile` — warns about orphan concepts but doesn't delete them
3. `rm wiki/concepts/<orphan>.md` for each one

**Concrete example:** ingesting an X.com post without auth captured just the error page:

```
$ cat sources/untitled.md
---
source: https://x.com/himanshustwts/status/2039811786602607052
---
Something went wrong, but don't fret — let's give it another shot.
Some privacy related extensions may cause issues on x.com. Please disable them and try again.
```

llmwiki then compiled 3 concept pages from that error text (`xcom-generic-error-page`, `privacy-extension-compatibility-with-xcom`, `browser-extension-interference-with-web-applications`), all of which had to be deleted by hand after removing the source.

**Proposal:** a single command that does all three steps.

```bash
llmwiki rm <source>       # delete source + derived concepts, then recompile
llmwiki rm --dry-run ...  # preview
```

Implementation can read the source→concepts mapping that already exists in `.llmwiki/state.json`.

**Alternative:** a `compile --prune-orphans` flag would also work.

PipDscvr:
Closes #60.

Adds `llmwiki rm <source>` — one command to undo a bad ingest, replacing the three manual steps the issue describes.

```bash
llmwiki rm <source>              # apply
llmwiki rm <source> --dry-run    # preview; takes no lock, changes nothing
```

It deletes the source file, hard-deletes the concept pages derived **exclusively** from it, preserves pages another live source still contributes to, updates `state.json`, and regenerates the index, MOC and embeddings. Broken wikilinks and pending review candidates are **reported, not repaired** — `rm` doesn't edit pages you didn't name.

## Decisions worth your sign-off

**Hard-delete, not orphan.** Your comment said "remove **or** orphan" without choosing. This removes, because the framing was that `rm` should collapse all three manual steps — and step three was `rm wiki/concepts/<orphan>.md`. Orphaning would leave that step behind. Swapping to `markOrphaned` is a small change if you'd rather.

**No confirmation flag.** The surface is exactly the two lines you specified. Note this is a weaker guard than `llmwiki state reset`, which requires `--yes` for an operation recoverable from its own `.bak`. Mitigated by `--dry-run` and by journalling the page deletes — but it's your call.

**Filename only, no URL.** "The matching file under `sources/`" read as a filename. Accepting a frontmatter `source:` identity is a small follow-up — `findFileBySourceIdentity` already exists privately — deliberately out of scope here.

## Preserving shared concepts

The concern you raised is enforced two ways rather than by convention:

- The exclusive/shared split delegates to `findSharedConcepts` — the same function `markOrphaned` uses — so `rm` and `compile` can't drift on that rule.
- The plan is computed lock-free (so `--dry-run` never takes the lock), then **re-verified against freshly-read state under the lock** before anything is deleted. If a concurrent `compile` makes a doomed concept shared in that window, it's dropped from the delete set. This follows the convention in `src/trust/executor.ts:178-181` that the under-lock handler is the authority and a caller-supplied plan is intent only. The intersection can only ever shrink.

A removal also **freezes** the concepts it still shares with a live source, matching what `findFrozenSlugs` does on compile's deletion path. Without that, a later recompile of a remaining contributor would silently drop the removed source's contribution to a page that was correctly preserved on disk.

## New: journalled page deletion

Nothing in the codebase deleted a wiki page before — every page mutation went through the journalled *write* executor, and `MutationOperation`'s `"delete"` variant was declared but never produced.

`src/wiki/delete-page.ts` composes the existing public journal API (`openBatch` → `recordPreState` → `confinedUnlink` → `commitBatch`), so a crash mid-removal replays the pages back. This deliberately avoids extending the trust executor: the only change to `src/trust/journal.ts` is exporting the already-private `confinedUnlink` plus a JSDoc block.

Every unlink is **verified, not trusted**. `confinedUnlink`'s catch is shared with the revert path, where swallowing is correct (absent *is* the goal there); reused for a delete it would also swallow `EACCES`/`EROFS`/`EBUSY`, reporting success for a page still on disk. `assertUnlinked` re-checks with `lstat` and throws before commit, leaving the batch pending for replay.

## Two things you should know before merging

**`rm` is now a reserved core verb.** Required by `test/reserved-verbs-drift.test.ts`, which enforces that no profile-declared workflow id can shadow a CLI verb. **This is a backward-compat break:** a profile declaring a workflow keyed `rm` now fails validation and must rename it. Called out in the CHANGELOG.

**Limited value on profile projects.** Typed entity pages record no source ownership — `review-approve.ts` deliberately skips `persistCandidateSourceStates` for candidates with a `targetEntityType` — so `rm` can't find or remove them, and neither can `compile`'s orphan pass. On a profile project `rm` still removes the source file and any `wiki/concepts/` pages, and now prints a warning naming the profile and stating the limitation. It doesn't refuse, since a profile project can legitimately have concept pages. Documented as a known limitation. Recording that ownership is the real fix and belongs in its own change — happy to open an issue.

## Found while building — reported, not fixed

`confinedUnlink`'s bare `catch` means the journal's **revert** path can silently fail to restore a pre-state during crash recovery. Pre-existing, and out of scope here to keep the trust-layer diff to one `export`. `assertUnlinked` in this PR is the pattern the revert path could adopt. Happy to open a separate issue.

## Testing

```bash
npm run build
npm test
```

Suite: **4447 passing, 0 failures** — 26 new tests across the delete primitive, the planner, resolution, the CLI, and end-to-end.

One environment note that cost me some time and may be worth a line in CONTRIBUTING: six tests asserting "fails with a credential error when no API key is set" fail on any machine with an active Claude Code login, because `claude-agent` requires no credential env var (`provider-guard.ts:45`) and authenticates against that session. Running `LLMWIKI_PROVIDER=anthropic npm test` with no `ANTHROPIC_API_KEY` restores them. CI is unaffected.

To exercise it by hand:

```bash
llmwiki rm <source> --dry-run   # nothing changes, no lock taken
llmwiki rm <source>             # applies
```

`npx tsc --noEmit`, `npm run build`, `npm run fallow:ci` (0 above threshold) and `npm run release:check-docs` all pass.

## Docs

New `docs/cli/rm.mdx` with nav entry, a README core-commands row, and a CHANGELOG entry. The page documents what's kept vs deleted, the exit codes, the profile limitation, and the confinement / journalling / lock-refusal boundaries.


ethanj (changes requested):
Requesting changes on two things, both with small fixes. Everything else here is decisions or notes.

First the part that isn't in question: the confinement work is genuinely good. Unlink routes through `confineUnderRoot`, which walks the ancestor chain rather than checking the leaf, so a symlinked parent is caught — and there's an adversarial test asserting a file outside the project survives. `rm` goes into the one `RESERVED_CORE_VERBS` enumeration rather than a parallel list. `readStateFailClosed` refusing instead of falling back to the fabricate-empty path is exactly right, and the docstring explains why. `assertUnlinked` re-`lstat`ing after every unlink, so a swallowed `EACCES` can't be reported as success, is the discipline I'd want on a delete. Reusing `findSharedConcepts` rather than reimplementing the split means `rm` and `compile` can't drift.

## 1. Blocking — the under-lock recheck filters the stale plan instead of recomputing ownership

https://github.com/atomicstrata/llm-wiki-compiler/blob/184459d692c8235c9423ad8cc4f2392c9a75f340/src/sources/removal.ts#L234-L255

https://github.com/atomicstrata/llm-wiki-compiler/blob/184459d692c8235c9423ad8cc4f2392c9a75f340/src/sources/removal.ts#L292-L298

State is re-read under the lock, but `reverifyDeletable` only asks, of the slugs already in `plan.deleteSlugs`, which are now shared. `findSharedConcepts` iterates only `state.sources[sourceFile].concepts`, so anything that has left that list is never evaluated. The recheck therefore proves "not newly shared" when the property it needs is "still ours."

Three directions get through, and they are not equally bad:

- **A concept transferred away from this source keeps its stale delete-verdict.** A concurrent compile moves `X` from `alpha.md` to `beta.md`; `X` drops out of alpha's fresh list, so it never enters `sharedNow`, so it stays deletable — and `wiki/concepts/X.md`, now owned exclusively by the untouched live source `beta.md`, is deleted. `deleteWikiPagesLocked` does no ownership check of its own. Worst case of the three: silent loss of a page belonging to a source nobody named.
- **A concept that became exclusive stays in `keptSlugs`**, survives, and ends up owned by nothing.
- **A concept newly added to this source isn't in the plan at all**, so its page is left behind untracked.

The extreme case is worth stating because the code allows it: `findSharedConcepts` opens `if (!sourceEntry) return shared;`. If a concurrent compile has already dropped the source's state entry, `sharedNow` is empty and *every* slug in the stale plan is deletable.

This isn't hypothetical scheduling. `llmwiki watch` auto-recompiles on changes under `sources/`, so the trigger is the documented workflow: `watch` running in one terminal, `rm` in another, plan computed lock-free while a debounced compile is mid-flight.

**Suggested fix, deliberately small:** don't recompute an expanded delete set under the lock. Compare the source's fresh ownership against what the plan assumed, and if anything moved, abort and tell the user to re-run. Refusing on a changed world is simpler than reconciling it, and on a destructive path it's the right default. The existing race test only covers the additive case — it pushes a slug onto *another* source's list while leaving the doomed source's list intact — which is exactly why this survived.

## 2. Blocking — the source is deleted before the fallible part, so a failure can't be retried

https://github.com/atomicstrata/llm-wiki-compiler/blob/184459d692c8235c9423ad8cc4f2392c9a75f340/src/sources/removal.ts#L244-L249

`deleteSource` runs first, then the journalled page batch, which can throw — `assertUnlinked` is *designed* to throw when a page survives its unlink. When it does, the source file is already gone. Re-running the command doesn't work:

https://github.com/atomicstrata/llm-wiki-compiler/blob/184459d692c8235c9423ad8cc4f2392c9a75f340/src/sources/removal.ts#L75-L83

`resolveSourceRef` resolves through `getSource`, so an absent source file is "no such source" and the second `rm` exits before it reaches anything. The user runs the command, it fails, they run it again, and it tells them the thing they were removing doesn't exist — while its pages are still on disk and the batch is still pending.

This also makes one documentation line wrong. `docs/cli/rm.mdx` lists "the next `rm`" alongside `compile` and `recover` as ways a pending batch gets replayed. `compile` and `recover` do work, because neither needs the source file; `rm` is the one that can't, and it's the one a user will reach for first.

**Suggested fix, also bounded:** let `rm` resume when the source file is already absent but a state entry for it still exists. That's the case that means "a previous `rm` got partway," and it's distinguishable from "you typed a name that was never there."

## 3. Non-blocking — a preserved page can quietly end up with a broken citation

`computeRemovalPlan` reports the two kinds of collateral it knows about: surviving pages whose `[[wikilink]]` points at a doomed slug, and pending candidates referencing the source.

https://github.com/atomicstrata/llm-wiki-compiler/blob/184459d692c8235c9423ad8cc4f2392c9a75f340/src/sources/removal-plan.ts#L104-L118

It doesn't look at citations. A shared page that is correctly preserved can still carry `^[bad.md:1]` pointing at the source just deleted, and the next `lint` raises a broken-citation error the `rm` transcript gave no hint of. Given the command's whole stance is "report integrity damage, repair nothing," provenance damage is the one category missing.

I don't think this needs citation analysis inside `rm`. A flat line whenever any page is preserved — surviving pages may still cite this source, run `llmwiki lint` — closes the gap at the right cost.

## 4. Minor — a documented output that can't happen

`docs/cli/rm.mdx` promises `Not deleted: <slug> (floor:unsafe-slug)` with exit 1. The floor exists in `delete-page.ts`, but every concept slug in state has already passed `isSafeFilenameComponent` during state validation (`src/utils/state.ts:161`), so a state carrying an unsafe slug fails closed as corrupt before a plan is ever produced. The floor is fine as defence in depth; the promise of user-visible output isn't reachable. Either drop that sentence or say it's unreachable by construction.

## Decisions, not defects

**Hard-delete vs orphan.** My comment on #60 said "remove *or* orphan" and never chose — that ambiguity is mine, not yours. Delete is the right read given the whole point was collapsing the third manual step. Keep it.

**No confirmation flag.** You flagged that this is a weaker guard than `state reset`, which requires `--yes` for something recoverable from its own `.bak`. That's true, and the defaults are in fact inverted — `state reset` previews unless told otherwise, `rm` applies unless told otherwise. But the command is named `rm`, `--dry-run` exists, and the two-line surface is the one I specified. I'm content to ship it as-is; noting it so the choice is on the record rather than defaulted into.

## One thing to coordinate

#179 writes shared-but-doomed slugs into `state.frozenSlugs` and documents those pages as kept untouched. #171 is open and rewrites what that field means, so a frozen shared page *is* rebuilt on the next compile with the removed source's contribution dropped — its own test asserts the rebuilt body no longer contains the removed source. Whichever of these lands second silently voids the other's guarantee. Not a defect in either, but it needs settling before both are in.

Items 1 and 2 are what I'd like fixed; 3 and 4 are small enough to take in the same pass if you're in there anyway. This is careful work — the failure mode in item 1 is one you'd already built the machinery to prevent, and it only covers one direction of it.

PipDscvr:
Both blocking items are fixed in d0daf8b, along with 3 and 4. Below is what changed, one thing I got wrong about the window's shape, and a proposal on the #171 overlap.

## 1 — ownership is recomputed, not filtered

Confirmed before changing anything, running the old re-check against real `findSharedConcepts` with `race` transferred from `bad.md` to `good.md`:

```
plan.deleteSlugs   = [ 'junk', 'race' ]
sharedNow          = [ 'shared' ]
OLD deletable      = [ 'junk', 'race' ]   <- 'race' is now good.md's only page
NEW current.delete = [ 'junk' ]           <- drift detected
```

The split now lives in one function, `partitionConcepts(sourceFile, state)` in `removal-plan.ts`. `computeRemovalPlan` calls it against the pre-lock read; `applyRemovalLocked` calls it against the under-lock read. `assertOwnershipUnchanged` compares **both** halves as multisets and throws before the first mutation if they differ.

I took the refusal, not reconciliation, for the reason you gave — and the two are the same function now, which is the part that actually closes this. The previous code computed the plan's split with one expression and re-verified with a different one; they were never comparable, only individually plausible.

Consequences worth flagging:

- **`RemovalApplyResult.preserved` is gone**, along with the `(became shared with another source during removal)` line. With refusal there is no mid-race preserve.
- All four drifts are now pinned as tests, each asserting the source file, the exclusive page, and the shared page are *all* still on disk: newly-shared, **transferred away**, **state entry dropped**, and the additive case that was already covered.

**One correction to the risk picture.** I stated in the PR body that the plan/apply gap spans dry-run to apply. It doesn't — `--dry-run` and a bare `rm` are separate processes, and the bare `rm` computes its own plan. The window is intra-process only, between `planRemoval` and `acquireLock` inside one invocation. It's still real and `watch` can land in it, and the transfer case was a genuine silent-data-loss path, but the refusal will fire less often than "a dry-run went stale" would suggest. I've said so precisely in the docs rather than overstating it.

## 2 — an interrupted removal can be resumed

`planRemoval` now also resolves a source whose file is absent but whose `state.sources` entry survives.

I made the absence check narrower than `getSource(...) === null`, because that also covers present-but-invalid — a symlink, a directory. Only genuine `ENOENT` means "a previous removal got partway", so there's a new `sourceFileMissing` in `store.ts` using `lstat`. A symlink occupying the name stays a plain "no source matches" even when state carries an entry for it (pinned as a test) — `rm` should never be resumable *around* a path it refuses to treat as a source.

The transcript follows: `RemovalPlan.sourcePresent` and `RemovalApplyResult.sourceDeleted` mean a resumed removal prints

```
i sources/research-notes.md was already gone — only its pages and state entry remain
```

instead of claiming to delete a file that was already gone. That's the one place a false `Deleted:` line would actively mislead someone mid-recovery. `deleteWikiPagesLocked` already replays the pending batch first, so the retry restores then re-deletes as one clean batch.

Two knock-on notes: the `docs/cli/rm.mdx` line about "the next `rm`" is now true, and written up under a *Resuming an interrupted removal* section. And this incidentally settles the "can't `rm` a source whose file is already deleted" limitation I'd left for you — it now works, and I think that's right rather than an accident.

`resolveSourceRef` is replaced by a pure `toSourceId` (validate + `.md`, no I/O), since "absent" and "unsafe" can no longer collapse to one `null`. It wasn't part of the SDK surface. Its tests now exercise resolution through `planRemoval`, where the resume rule is actually observable.

## 3 — citations

Added, as the flat line you specified:

```
! Kept page(s) may still cite this source.
Run `llmwiki lint` to find any citations left pointing at it.
```

Gated on a page actually being kept — a deleted page takes its citations with it. Not a citation scan, for the reason you gave: it would duplicate `splitCitationMarker` / `stripSpanSuffix` handling on a destructive path to report what `lint` already reports exactly.

## 4 — the unreachable floor output

Said unreachable rather than dropped, since the mechanism is worth explaining:

> That warning is unreachable by construction, and is kept only as defense in depth. Every slug `rm` can reach comes from `.llmwiki/state.json`, and reading that file already rejects an unsafe slug — the whole state is classified corrupt, and `rm` refuses before a plan exists.

Exit-code table marks the row accordingly. The floor itself is untouched.

## On #171

I think these compose without either voiding the other, but it needs your call.

`rm` writes into `state.frozenSlugs` through the field `findFrozenSlugs` already maintains — it isn't inventing a meaning, it's reusing compile's. What #179 actually depends on is that the slug gets **marked as having lost a contributor**; without the mark, the next recompile of a surviving contributor rebuilds the page from live sources as if nothing had been removed. What compile then *does* with a marked slug is compile's policy: today `mergeExtractions` skips regeneration; under #171 it reconciles through the owner closure and drops the removed source's contribution with citations filtered. Both need the mark; only #171's is the better answer.

So I've removed the forward-looking promise from #179 rather than defending it — the docstring and docs now say the freeze marks the page and defers the policy to compile, and the table row reads "Kept, and marked as having lost a contributor" instead of "Kept, untouched". The CLI's `Kept: … (shared with other sources)` line was already neutral.

Net: if #171 lands second it needs no change here, and if #179 lands second #171 inherits `rm`-set marks unchanged. If you'd rather `rm` not write `frozenSlugs` at all and leave the whole question to #171's reconciler, say so and I'll pull it — but then a removal followed by a recompile silently loses the contribution until #171 lands, which is why I left it in.

## Not changed

Hard-delete and the no-confirmation surface, both as you decided. `--dry-run` remains the only pre-flight.

## Verification

`4458 passed / 0 failed` (620 files), `tsc --noEmit`, `npm run build`, `fallow` 0 above threshold / no dead code / no duplication, `release:check-docs` clean.

One note for `CONTRIBUTING.md` if you want it: the suite needs `LLMWIKI_PROVIDER=anthropic` with no `ANTHROPIC_API_KEY` on a machine where Claude Code is logged in. `claude-agent` requires no credential env var (`provider-guard.ts:45`) and authenticates from `~/.claude`, so six tests asserting a missing-credential error find working auth and fail. It also blocks the pre-push hook.

ethanj (approved):
This is the version I wanted, and the ownership fix is better than what I asked for.

I suggested comparing fresh ownership against what the plan assumed and refusing if anything moved. You did that and then went a step further, extracting `partitionConcepts` so the pre-lock plan and the under-lock check compute the split with one function instead of two. That is what makes the guard total rather than partial: the old check asked which of the planned slugs had become shared, which can only ever see one of the ways ownership moves, while comparing the whole partition catches every direction by construction. Your docstring puts it better than I did, and the three race tests exist because the fix made them expressible.

The simplification that fell out is the tell. `preserved` is gone from the result, because once the run either proceeds against an unchanged world or refuses outright, there is no mid-removal category left to report. A fix that deletes a concept usually landed in the right place.

`sourceFileMissing` is careful in exactly the spots that matter. `lstat` rather than `stat`, so a dangling symlink reads as present and never gets mistaken for an interrupted removal, and narrower than "getSource returned null" so a directory or an unreadable entry at that path is not silently deleted around. Making the docs line about resuming true, rather than deleting the line, was the better of the two options.

The citation warning is scoped the way I would have scoped it, and gating it on a kept page existing because a deleted page takes its citations with it is the part I would not have thought to write down.

Landing it.

Issue #178, opened 2026-08-12
feat(viewer): make CLP entities drillable to provenance, relations, and artifacts


ethanj:
## Context

Follow-up to #176 and #177.

PR #177 makes CLP entity types visible in `llmwiki view`: typed pages are listed, searchable, routable, and represented in the graph and pipeline. It does not yet expose the information that connects those records to original sources, related entities, or stored artifacts.

The result is structurally browseable but not meaningfully drillable. For example, an AutoSci paper page can contain authors, year, DOI, arXiv ID, lifecycle state, connector provenance, citations, and typed relations, while the viewer shows only its Markdown body and a generic metadata rail.

## Goal

Generate a read-only CLP entity experience from the active profile and instance data, without adding domain-specific branches for AutoSci, Newsroom, or future profiles.

A reader should be able to move through:

`source → entity → related knowledge → experiment → artifact → output`

The viewer must never infer or invent a connection that is not present.

## Existing data to reuse

CLP already declares or stores most of the required information:

- Entity fields, field types, required fields, enums, and `titleField`
- Lifecycle field, states, transitions, and transition requirements
- Typed relation definitions and live relation instances
- `artifactRef` fields, artifact types, content kinds, metadata, and health
- Durable connector provenance, including the fetched source URL
- Claim citations and source spans
- Workflow definitions and entity/artifact dependencies

### 1. Project the relevant profile schema to the viewer

Add a bounded `profileSchema` read model for non-default profiles containing:

- Entity type, directory, `titleField`, and declared fields
- Field type, required status, enum values, and artifact scopes
- Lifecycle field and states
- Relation endpoint definitions and direction
- Artifact type, content kind, filename, and metadata schema

Do not send executable behavior, absolute paths, or unrelated profile configuration.

### 2. Render entity metadata generically

On typed entity pages:

- Render every declared field present on the record
- Use the declared `titleField`
- Render enum/lifecycle values as state labels
- Render arrays as lists
- Render dates, numbers, and booleans according to their declared types
- Render `artifactRef` values through the artifact resolver
- Distinguish missing optional values from invalid or unresolved values

Do not hardcode field names such as `doi`, `authors`, `hypothesis`, or `verdict` in viewer code.

### 3. Add per-page relation drilldown

Include live incoming and outgoing relations in the page payload.

[...]

PR #180, opened 2026-08-13
feat(profile): title a typed page by its declared titleField

PipDscvr:
Prerequisite for #178 Slice A, split out per [your ruling](https://github.com/atomicstrata/llm-wiki-compiler/issues/178#issuecomment-5274341115): a PR whose subject is the title move, landing before A so A inherits it.

## What was wrong

`EntityTypeDef.titleField` has been in the profile schema since the format shipped and was read by nothing. `pageFromScan` took the literal `title` key, so a type carrying its display name under another key had no title at all and every surface fell back to the slug. AutoSci's `people` is exactly that type — it declares `name`.

It was also unvalidated: schema-typed as a bare `{"type": "string"}` and compared against nothing, so a profile could declare `titleField: "headline"` against a type with no `headline` field and load clean. Pages would then show their slug with no diagnostic on any surface — the failure the field exists to prevent.

## The fix

**Resolved in the collector, not in a reader.** `status`, JSON and OKF export, context packs, index generation, lint and the viewer now read one declaration one way. I had originally proposed a viewer-local fix; you were right that it gives a single declaration two meanings, which is the dual-source problem this codebase keeps removing.

**Narrow by construction.** A type declaring no `titleField` keeps the previous path byte-for-byte, `parseStatus.hasTitle` included. An unusable value — absent, blank, or not a string — yields `undefined`, so every existing slug fallback behaves as before. The declared path additionally trims, so a title of `"   "` is treated as absent rather than rendered as a blank heading; the existing path was left exactly as it was, and the asymmetry is commented as deliberate.

**Validation.** `titleField` must name a declared `string` field. `slug` is excluded on purpose — the slug is already what an untitled page falls back to, so titling by one would be a no-op dressed as a choice. `string[]` too: a title is one value, and picking an element would be a rule the profile never stated.

## AutoSci 0.2.0, and why 0.1.0 is retained

All twelve types declare `titleField` (`people` by `name`); the template and profile advance to `0.2.0`.

Retention turned out to be load-bearing rather than ceremonial. `planTemplateUpdate` resolves the **installed** release and compares its `profileDigest` against the on-disk profile to decide whether a project carries local modifications. So:

- dropping `0.1.0` makes `planBuiltinTemplateUpdate` throw for every project still on it — it fails closed with `cannot independently resolve installed builtin`;
- a retained `0.1.0` that had silently gained `titleField` would report every genuine `0.1.0` install as locally modified.

`0.1.0`'s entity block is therefore **derived** from `0.2.0` by dropping the title declarations — the only difference between the releases — rather than frozen as a second 190-line copy that would be a second thing to keep correct. The derivation is not taken on trust: `test/profile-template-autosci-releases.test.ts` pins its `profileDigest` to `bf48a4e7…`, which I computed from the pre-change source rather than from the derivation, so it is an independent check. A later edit to the `0.2.0` entities that leaks into the derived `0.1.0` fails there instead of quietly mis-describing an installed project.

`isShippedBuiltinProfile` now also recognises `0.1.0` bytes, so a project still on it reads as shipped rather than modified.

## No golden moved — and that is itself a finding

You expected goldens to move. None did, and the reason is worth stating: the parity goldens cover the default profile, and **nothing in the suite exercised a page whose title field is not `title`**. Every AutoSci type except `people` already declared a `title` field, and no test covered a `people` page's title. So the one type this change exists for was the one type with no coverage.

`test/profile-title-field.test.ts` adds it: the declared field is read; a literal `title` is ignored when the type titles by another field; the omitted-`titleField` path still reads `title`; and absent / blank / non-string values all leave the title undefined so the slug fallback is unchanged.

Three template tests did break, all on a hardcoded `0.1.0` literal. They now read the version from the registry, since they are about install mechanics rather than about which release is current — that fact is pinned once, in the releases test where it is the subject.

## Docs

`docs/configuration/profiles.mdx` gains a **Page titles** section: what `titleField` names, that it must be a declared `string` field, which surfaces consume it, and that adding it to an existing profile changes what those surfaces display.

## Verification

`npx tsc --noEmit` clean · `npm run build` clean · **4882 passed, 3 skipped, 0 failures** · `npm run fallow:ci` clean.

Run the suite as `env -u LLMWIKI_PROVIDER -u LLMWIKI_MODEL npx vitest run` — the six credential tests read ambient env.

## Not included

No `format` vocabulary, no `profilePipeline` extension, no renderer. Those are Slice A. AutoSci takes a second bump to `0.3.0` there for the `doi`/`arxiv`/`url` formats, since `format` cannot be declared before the schema accepts it — flagged in the issue thread rather than folded in here ahead of the schema that makes it valid.


PipDscvr:
Pushed `4940e2f` — **newsroom needed this more than AutoSci did, and my first pass missed it.**

None of newsroom's three types declares a `title` field at all: `articles` carries `headline`, `desks` carries `name`, `bylines` carries `reporter`. So it wasn't one type showing a slug, it was **every newsroom page on every surface**. AutoSci's `people` was the visible case; newsroom was the whole template.

That also matters past the template itself — newsroom is the second profile `no-research-branch-in-core` uses, and the fixture #178's acceptance criteria name. Landing a `titleField` fix that left it slug-only would have made that comparison meaningless.

Same treatment: a declaration per type, `0.2.0`, `0.1.0` retained with its published digest (`8668828…`, computed from the pre-change source) pinned.

**The derivation is now shared.** Both templates needed the identical "current minus the title declarations" transform, so it lives in `src/profile/templates/title-fields.ts` — deliberately **outside** `builtin/`. That directory is the template-data allowlist `profile-template-genericity` enumerates, and the guard failed when I first put the helper there. It was right to: the helper names no vocabulary, and adding it to the data allowlist would have blunted what the allowlist means. Moving it out kept the guard green and unedited.

The two release tests are now one table-driven file over both templates, since the contract is identical and two copies of it would drift.

`tsc` / build / **4892 passed, 3 skipped, 0 failures** / `fallow:ci` all clean.

One consequence for the PR body above: the *"no golden moved"* note still holds, but the reason is now sharper. Nothing in the suite asserted a typed page's title, for either shipped profile — so the two templates whose pages were entirely untitled were also the two with no coverage of the thing that was broken. `test/profile-title-field.test.ts` is what closes that.

PipDscvr:
**Correction — I was wrong to call the `empty-page` regression refuted, and this PR shipped it.**

A review of the follow-up work flagged that resolving `titleField` in the collector would newly arm `empty-page` for frontmatter-only records. I checked `src/linter/rules.ts`, found it derives the title from `meta.title` directly, and reported the finding as refuted.

There are two linters. `src/profile/lint.ts:94` passes `EntityPage.title` — the field this PR changed — straight into `checkPageEmpty`. Measured on the shipped newsroom profile with a frontmatter-only `desks` record:

```
BEFORE titleField: []
AFTER  titleField: ["empty-page"]
```

So adopting newsroom `0.2.0` or AutoSci `0.2.0` makes every `desks`, `people`, `bylines` and `topics` record emit a lint warning — record types whose normal shape is fields with no prose — and drags down `llmwiki eval`'s health score and the viewer's header verdict with it. That is a real regression in this PR, not in the follow-up.

I checked one file, found the answer I expected, and stopped. The failure was treating a single negative grep as proof of absence for a field with a dozen consumers.

**Fixed** on the follow-up branch, along with a second consumer the same audit turned up:

- `empty-page` now reads the **literal** `title` frontmatter key. The rule asks whether a page that *announces* a title has prose beneath it, and the announcement is that key — not whichever field the type displays by.
- `applyEntityStandardFields` was writing the resolved title into **OKF frontmatter**, so a `desks` record gained a `title` key it never had while `collectDomainFields` — which suppresses only a field literally named `title` — kept exporting `name`. One value under two keys, in a published interchange format, invisible to `contentHash` because that is computed over the body. It now carries what the page wrote; the resolved title stays where a *display* title is wanted (the bundle TOC, the viewer, the graph).

`test/profile-title-field-blast-radius.test.ts` pins both directions, including that a record which *does* write a literal `title` and has no prose is still flagged.

Since the fixes live on the follow-up branch and this PR is the one that introduced the regression, tell me which you'd prefer: I can cherry-pick both onto this branch so it lands clean, or leave them stacked and you take this one knowing the follow-up closes it. My preference is cherry-picking — a PR that ships a lint regression and a format widening shouldn't rely on the next one to fix them.

ethanj (approved):
This is excellent work.

The part I want to call out is not the feature, it's the blast radius. Resolving `EntityPage.title` through `titleField` would have quietly broken two things, and you caught both: `empty-page` would have started flagging every frontmatter-only record type, turning "adopt a shipped template" into a health-score regression for a project that did nothing wrong; and OKF would have shipped one value under two keys, since `collectDomainFields` suppresses only a field literally named `title`. Keeping both on the announced key, with the reasoning written down at each site, is the right call and not an obvious one.

I checked the `0.1.0` digests against the pre-change source rather than taking the pin on faith, and both templates reproduce exactly. That makes the retained releases a real check on the derivation instead of a restatement of it, which is the difference between a test that would catch a leak and one that would ratify it.

"No golden moved, and that is itself a finding" is the other thing worth noting. The one type this change exists for was the one type with no coverage, and saying so plainly is more useful than a green run.

Landing it. Thanks for splitting this out ahead of Slice A.

PR #189, opened 2026-08-21
feat(viewer): render a typed entity page's declared fields and link formats

PipDscvr:
Slice **A** of #178, per the A/B/C split agreed on that issue. Refs #178 — it does not close it; B (relations, provenance) and C (artifacts, the two source surfaces) follow.

A typed entity page used to show its body beside a rail describing a contract it was never under: the rail's nine fixed keys are the *default* profile's vocabulary, so a paper's authors, year, DOI and stage were simply invisible. This makes the page render what its own type declares.

## What's in it

**§1 — the read model.** `profilePipeline` gains `titleField` and `fields`, plus a sibling `artifactTypes`. No `profileSchema` block: `type`, `directory`, `lifecycle` and the relation endpoints already ship in `profilePipeline`, and a second block able to disagree with it about a lifecycle is the shape we've spent three PRs removing. The name is kept, with a docstring stating the block's remit.

**§2 — generic rendering.** The renderer dispatches on the declared `FieldDef.type` and never on a field name; `test/no-research-branch-in-core.test.ts` stays green and unedited. Three value states are reachable and no more — present, absent-and-optional (row omitted), and unresolved artifact reference — because a record violating its declared field contract never becomes a viewer page: `collectTypedViewerInputs` filters it and it surfaces as a `field-violation` problem instead. There's a comment at the branch point saying why there is no fourth state.

**§6 — declarative link formats.** A closed `url | doi | arxiv` enum, no author-supplied URL templates. Both layers: an unknown format fails validation, *and* the renderer falls back to plain text for anything unknown or malformed rather than trusting that a validator ran upstream. Origins are out of author control entirely — `url` links only `http:`/`https:`, `doi` through a fixed `https://doi.org/`, `arxiv` through a fixed `https://arxiv.org/abs/`.

**AutoSci** declares `doi`, `arxivId` and `locator` formats, taking it to `0.3.0`. `0.1.0` and `0.2.0` are both retained as resolvable releases, derived backwards from the current pack — `0.2.0` is `0.3.0` without the formats, `0.1.0` is that without the title declarations — with each published `profileDigest` pinned in `test/profile-template-releases.test.ts`, so a later edit that corrupts a derivation fails there rather than quietly mis-describing an installed project.

## Digest drift, and what it is not

Adding `format` to a `FieldDef` shifts the profile and template digests. It does **not** reclassify in-flight workflow runs: `classifyRun` compares `workflowDefDigest(def)` against `run.workflowDigest` (`src/workflows/status.ts:181-189`), and `run.profileDigest` is pinned at start and shape-checked in `run-validate.ts:101` but read by no classification path. Calling this workflow adaptation would send people looking for an effect that isn't there.

## Nothing disappears from the rail

Checked rather than assumed, since it was the one open question on §2. `appendFrontmatterDl` iterates a fixed nine-key `RAIL_FIELDS` list and skips a key only when the declared block already rendered it. The rail never displayed arbitrary frontmatter, so a key that is both declared and in the nine *moves* between lists rather than being lost, and an undeclared key outside the nine was never displayed before and still isn't. No regression to place.

## One bug found reviewing my own work

`bfd45f9` is the same class of defect as the `titleField` lookup corrected in #187, in the two places this slice introduced it. The schema constrains a field's `type` to a closed enum but puts no `propertyNames` on `fields` or on the entity map — so the declared **name** is the open half, and every one of these lookups indexed a plain JSON object with one.

- `viewer-entity-fields.js` read `frontmatter[def.name]`. Frontmatter arrives as JSON off the wire, carrying `Object.prototype`, so a type declaring a field named `constructor` resolved `Object.prototype.constructor` on *every* page of that type and rendered `function Object() { [native code] }` as the field's value — a value the record does not carry, displayed as though it does. `toString` and `valueOf` behave the same.
- `pipeline.ts` joined counts onto declarations by indexing summary maps with a declared type name, so a type named `constructor` produced `expected [Function Object] to be +0` — and `JSON.stringify` then drops the key, so the row reached the client missing a `pageCount` it is required to carry.

Both are pinned by tests that fail without the fix. `FIELD_RENDERERS` already took the null-prototype form of this precaution for the declared *type*; this is the same precaution for the declared *name*.

**Worth your call separately:** I hardened the consumers, which is this slice's remit. The underlying gap is that the schema puts no `propertyNames` constraint on declared names at all. Constraining them upstream would touch profile validation and could reject profiles that currently load, so I didn't fold it in — but it's the root, and it will keep producing this bug class.

## Testing

```bash
npm run build
npx vitest run test/viewer-entity-fields.test.ts test/viewer-entity-fields-contract.test.ts \
  test/viewer-profile-schema.test.ts test/viewer-pipeline-envelope.test.ts \
  test/viewer-rail.test.ts test/profile-template-releases.test.ts
```

To see it: `llmwiki view` on an AutoSci project and open a paper — authors, year, stage and a working DOI link where there was a slug and a generic rail. The same renderer against the Newsroom fixture, with no domain-specific conditional.

Full gate: `tsc --noEmit`, `npm run build`, **4989 passed / 0 failed**, `fallow` 0 above threshold with no dead code or duplication.

One environment note: the suite needs `LLMWIKI_PROVIDER=anthropic` with no `ANTHROPIC_API_KEY` on a machine where Claude Code is logged in — `claude-agent` requires no credential env var (`provider-guard.ts:45`) and authenticates from `~/.claude`, so tests asserting a missing-credential error find working auth and fail. Same note as on #179.

## Next

**B** — §3 relation drilldown and §4 provenance, byte-free, reusing the relation instances already captured at snapshot creation, with the per-page list capped and the true total reported separately following `/api/reviews`. §4 links to the source *entry*, not the cited lines.

**C** — §5 artifacts and §7 the two source surfaces, written last so they sit at the top of the log: bytes on loopback only, non-loopback gets metadata and health, and the pinned hash re-verified at request time on every artifact response.

### DIFF
--- a/docs/cli/view.mdx
+++ b/docs/cli/view.mdx
@@ -131,6 +131,10 @@ For the per-count detail behind the verdict, open `#/health`. To narrow the conc
 
 **Page metadata.** Each page shows its frontmatter fields - kind, contributing sources, confidence score, provenance state, creation and update timestamps - in the metadata rail on the right of the page.
 
+**A typed entity page shows the fields its own type declares**, above that list and in the order the profile declares them. Values render by their declared type: an array becomes a list, an `enum` becomes a state chip, a boolean reads as Yes or No, and a field carrying a [`format`](/configuration/profiles#field-formats) becomes an external link. A field the record does not carry is left out rather than shown empty, and an `artifactRef` is named but marked unverified - the viewer does not yet read the artifact store to check it.
+
+Those fields lead because the list below them is the default profile's vocabulary and describes a contract a typed page was never under. Anything that list would have shown still shows, minus any key the profile declares as a field of its own - so nothing disappears and nothing is stated twice.
+
 **Freshness badges.** Pages whose underlying sources have changed since the last compile are labelled `STALE`. Pages whose sources were all deleted are labelled `ORPHANED`. Pages that declare contradictions in their frontmatter are labelled `CONTRADICTED`. Archived candidates show an `ARCHIVED` badge.
 
 **Provenance and citation chips.** Each paragraph's `^[source.md]` citation renders as a clickable chip. On loopback (`127.0.0.1`), citation chips include an editor link that opens the source file at the relevant line range directly in your default editor. Specific claim citations (`^[source.md:42-58]`) pin to the exact line range.
--- a/docs/configuration/profiles.mdx
+++ b/docs/configuration/profiles.mdx
@@ -76,6 +76,34 @@ Supported field types:
 Fields can be required with `required: true`, or by listing them in the entity's
 `requiredFields` array. Both forms are enforced on typed page writes.
 
+### Field formats
+
+`format` tells a read surface how to linkify a field's text. It is valid only on
+`string` and `string[]`, and takes one of three values:
+
+| Format | Renders as |
+|---|---|
+| `url` | The value itself, linked — only when it parses as an absolute `http`/`https` URL |
+| `doi` | A link through `https://doi.org/<value>` |
+| `arxiv` | A link through `https://arxiv.org/abs/<value>` |
+
+```json
+{ "doi": { "type": "string", "format": "doi" } }
+```
+
+The vocabulary is closed and the resolver origins are fixed — a profile names a
+resolver the reader already knows rather than supplying a URL template, so a
+profile can never point a reader at an origin of its choosing. An unknown format
+is rejected at load, and a value that does not match its format's grammar renders
+as plain text rather than as a link built anyway.
+
+<Note>
+  Adding `format` to an existing profile changes that profile's digest, and its
+  template's digest if it is a builtin. It does **not** affect in-flight workflow
+  runs: run classification compares the digest of the individual workflow
+  definition, not the profile's.
+</Note>
+
 ### Page titles
 
 `titleField` names the frontmatter key an entity type carries its display title
--- a/docs/configuration/profiles.mdx
+++ b/docs/configuration/profiles.mdx
@@ -97,6 +97,12 @@ profile can never point a reader at an origin of its choosing. An unknown format
 is rejected at load, and a value that does not match its format's grammar renders
 as plain text rather than as a link built anyway.
 
+Store a DOI as the identifier itself, without URL-encoding it first. The viewer
+encodes reserved characters such as `#`, `?`, `%`, and backslash as path data
+while preserving embedded `/` separators, so the resolver receives the complete
+identifier. If a value cannot survive URL normalization unchanged, it remains
+plain text.
+
 <Note>
   Adding `format` to an existing profile changes that profile's digest, and its
   template's digest if it is a builtin. It does **not** affect in-flight workflow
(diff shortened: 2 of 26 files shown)


ethanj:
Reviewing this as a draft, because early signal on 2000 lines is worth more than late signal. One thing to fix; everything else here is confirmation.

You honoured every ruling from #178, including the one where I overruled you: `titleField` resolves in the collector and the viewer projects only the declaration. `maxBytes` deliberately not projected is the one I was most worried about, since my §4 wording in the issue was loose enough to pull byte-serving into Slice A. You resolved that ambiguity the right way.

## One thing to fix: reserved characters truncate the DOI before it reaches the resolver

`DOI_PATTERN` accepts `\S+` on purpose, and the docstring is right that an over-tight grammar would degrade valid identifiers. But `new URL(value, base)` reads two of those permitted characters as component delimiters rather than as identifier text, so the resolver never receives the whole DOI:

```
10.1002/(SICI)1099-0518(199611)34:15<3129::AID-POLA9>3.0.CO;2-#
  href       https://doi.org/10.1002/(SICI)...3.0.CO;2-#
  pathname   /10.1002/(SICI)...3.0.CO;2-        <-- the "#" became an empty fragment
  server sees the truncated form, and doi.org 404s

10.1234/abc?x=1
  pathname   /10.1234/abc          search  ?x=1
  same truncation, via "?" instead
```

`<` and `>` are fine: the URL parser percent-encodes them and they survive in the path. It is specifically the two characters that open a new URL component. Crossref's own guidance on suffixes with special characters requires `#` to be sent as `%23`, and the first case above is their published example.

This is reachable rather than theoretical, since the Crossref connector preserves the DOI body as written.

The fix is to encode the reserved characters in the suffix while leaving the DOI path separators alone. Worth pinning Crossref's published example as a regression test, and `?` alongside it, since the two fail identically and a fix aimed only at `#` would leave the other one silently truncating.

Nothing else in the guard moves. The origin comparison is doing the real containment, exactly as your docstring argues, and it holds under everything I threw at it: `javascript:` with an embedded newline, an absolute `https://evil.test` smuggled into a DOI, protocol-relative arxiv, and formats named `__proto__`, `constructor` and `toString` all return null.

## Two things worth naming because they were not asked for

**The regression check is stronger than the one I specified.** I asked that nothing the rail shows today disappears. What you wrote drives the real projection into the real renderer, and opens with `renders a block at all, which a shape mismatch would silently prevent`. That is a vacuity guard on the rest of the suite, and it is the difference between a test that catches a projection and renderer drifting apart and one that passes while both do.

**You swept `own<T>()` across every profile-supplied name position** and cited `validateTitleField` for the reasoning. That is the same class of defect applied to entity types, relation types and lifecycle fields, which the profile schema leaves unconstrained. Finding the sibling surface of a bug without being asked is the part I would otherwise have had to request.

## Checked rather than read

The retained-release digests survived the rename to `prior-releases.ts`. I recomputed the new `0.2.0` pin from `main` independently and it matches, and `0.1.0` still carries the value verified during #180, so generalising the helper to N releases did not quietly loosen the control.

Both guards kill their mutants: passing everything through the link formatter turns 14 tests red, and letting format validation accept any field type turns 4 red.

It merges clean onto current `main` despite being three commits behind, with type-check clean and 116 tests green across the new and touched suites. Worth a rebase before you take it out of draft, since the CI green here is against `171d8895` rather than what it will land on.

Good work. Fix the DOI encoding and I think this is done.

PR #201, opened 2026-08-26
fix(cli): keep a rejected --concurrency value off stdout

### LINKED ISSUE #191: quickstart --json writes a warning to stdout before the JSON envelope on invalid --concurrency
### What happens

In `quickstart` JSON mode (`--json`), the command should write a single parseable JSON envelope to stdout. When `--concurrency` receives an invalid value (any non positive integer, for example `eight`, `0`, or `2.5`), a yellow ANSI warning line is written to stdout **before** the JSON envelope. The command still exits 0, so an automated caller reads stdout, tries to `JSON.parse` it, and fails on output that looks successful.

### Why it happens

`--concurrency` is parsed while the CLI builds the arguments to `quickstartCommand`, which runs before JSON mode enables quiet output.

- `src/cli.ts` evaluates `parseConcurrencyFlag(options.concurrency)` as an argument to `quickstartCommand(...)`.
- `parseConcurrencyFlag` (`src/compiler/concurrency.ts:44-53`) calls `output.status("!", output.warn(...))` for an invalid value, then returns `undefined`.
- `output.status` (`src/utils/output.ts:95-99`) writes to stdout via `console.log` unless quiet mode is on.
- Quiet mode is enabled later, inside `quickstartCommand` (`src/commands/quickstart.ts:146-148`, `output.setQuiet(jsonMode)`).
- `emitEnvelope` (`src/commands/quickstart.ts:586-590`) then writes the JSON to stdout.

Because the argument is evaluated before the function body runs, `setQuiet(jsonMode)` cannot suppress the warning, so stdout becomes `warning line + JSON`.

HEAD (default branch): https://github.com/atomicstrata/llm-wiki-compiler/blob/9c21d6d8734d5cd3e33e34b5c783e01652570170/src/compiler/concurrency.ts#L44-L53

### Reproduction

1. Create a Markdown source file.
2. Run `llmwiki quickstart source.md --json --no-open --concurrency eight`.
3. Capture stdout and `JSON.parse` the whole stream.
4. Parsing fails: an ANSI warning precedes the JSON object, even though the command exits 0.

### Failure scenario

Any tool or script that consumes `quickstart --json` output breaks silently whenever a user or wrapper passes a malformed `--concurrency` value. The exit code is 0, so the caller believes the command succeeded and instead receives unparseable data.

### Suggested fix

Either move `--concurrency` parsing inside the quiet-mode scope (parse after `output.setQuiet(jsonMode)` runs in `quickstartCommand`), or route the invalid-flag warning to stderr in JSON mode. Keep the existing fallback to the env var and default concurrency.

_Automated report: this issue was produced and filed automatically, with no human review before posting. Two independent checks agreed it is a real bug, but if it misreads the code please say so and we will close it._

_Found while running [Ito](https://ito.ai) (AI code review that runs your application, free for open source) against recently merged PRs. [Full analysis](https://app.ito.ai/share/eec41559-eac0-40c9-ad1c-79d37050fa46?tab=details&utm_source=oss-outreach&utm_medium=github-issue&utm_campaign=oss-autopilot)._

ethanj:
Fixes #191, reported by @graysoncooper with a diagnosis that pointed straight at the cause.

## What

`quickstart --json` promises a single parseable envelope on stdout. It did not deliver one when `--concurrency` was invalid:

```
$ llmwiki quickstart src.md --json --no-open --concurrency eight
! --concurrency value "eight" is not a positive integer; ignoring it.
{
  "version": 1,
  ...
```

Exit code is 0, so an automated caller sees success and then fails to parse.

## Why it happened

The value is parsed while the CLI assembles the command's arguments, at `cli.ts:389`, and quiet mode is enabled later inside the command body at `quickstart.ts:147`. Argument expressions are evaluated before the callee runs, so quiet mode was never early enough to suppress it. Moving the `setQuiet` call earlier cannot fix this, because there is no point inside the command that precedes its own arguments.

The channel was the real problem. stdout is the data channel, and a complaint about invalid input belongs on stderr. `output.note` already exists for exactly that and has eighteen callers, so the fix is to use it.

That also fixes `compile`, `refresh` and `watch`, which take the same flag, and it keeps the warning visible rather than trading a corrupt envelope for a silent one.

## Tests

- `test/quickstart-json-stdout.test.ts` drives the real CLI through `dist/` and asserts both halves of the contract: stdout parses whole, and stderr still names the rejected value. No credentials needed, since the provider failure is reported inside the envelope, which is the shape a caller has to read anyway.
- The existing unit test asserted the rejection appeared on stdout, so it encoded the defect. It now pins the channel split instead.

Both tests were mutation-tested against the committed fix. Restoring the stdout write turns both red, and deleting the diagnostic altogether also turns both red, so a fix that simply silenced the warning would not pass.

The two other warnings in `concurrency.ts` are left alone deliberately: they run inside the command, where quiet mode governs them correctly.

### DIFF
--- a/src/compiler/concurrency.ts
+++ b/src/compiler/concurrency.ts
@@ -38,6 +38,13 @@ function pickOverride(override: number | undefined): OverrideSource {
  * input so a typo like `--concurrency eight` reports "eight", not "NaN" — and
  * yields undefined so resolution falls through to the env var / default. A
  * valid integer is returned as-is; out-of-range clamping is the resolver's job.
+ *
+ * The rejection goes to stderr via `note`, not to stdout via `status`, because
+ * the CLI evaluates this while assembling a command's arguments, before that
+ * command's body can enable quiet mode. Anything written to stdout from here is
+ * therefore unsuppressible, and in a `--json` run it lands ahead of the
+ * envelope and breaks the parse for a caller that still sees exit 0. stderr
+ * keeps the diagnostic visible without touching the data channel.
  * @param raw - The flag value Commander parsed, or undefined when not passed.
  * @returns A positive integer, or undefined to defer to env/default.
  */
@@ -45,8 +52,8 @@ export function parseConcurrencyFlag(raw?: string): number | undefined {
   if (raw === undefined) return undefined;
   const n = Number(raw);
   if (!Number.isInteger(n) || n <= 0) {
-    output.status("!", output.warn(
-      `--concurrency value "${raw}" is not a positive integer; ignoring it.`,
+    output.note(output.warn(
+      `! --concurrency value "${raw}" is not a positive integer; ignoring it.`,
     ));
     return undefined;
   }
--- a/test/compile-concurrency.test.ts
+++ b/test/compile-concurrency.test.ts
@@ -21,19 +21,25 @@ import {
   ENV_COMPILE_CONCURRENCY,
 } from "../src/utils/constants.js";
 
-/** Run fn while capturing console.log (output.status writes there) and return the joined text. */
-function captureStatus(fn: () => void): string {
-  const logs: string[] = [];
-  const spy = vi.spyOn(console, "log").mockImplementation((m?: unknown) => {
-    logs.push(String(m));
+/** Run fn while capturing one console channel, and return the joined text. */
+function capture(channel: "log" | "warn", fn: () => void): string {
+  const lines: string[] = [];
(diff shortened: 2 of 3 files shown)


ethanj:
Superseded by #204: same stderr routing, plus coverage of all four rejection shapes, exit-code and envelope-shape pins, a committed human-mode witness, and doc updates.

PR #202, opened 2026-08-27
test: broaden quickstart --concurrency witnesses

ethanj:
## Broaden the quickstart --concurrency witnesses (stacked on #201)

Test-only additions to `test/quickstart-json-stdout.test.ts`, pinning the coordinates the
defect report named but the original oracle left open:

- **Input-space coverage**: parameterized over `eight` / `0` / `2.5` / `-1` — one case per
  rejection branch of `!Number.isInteger(n) || n <= 0`.
- **Exit-code pinning**: `expectCLIExit(result, 0)` in every case. `runCLI` never throws, so
  previously a nonzero-exit regression with an intact envelope would pass; the defect was
  precisely "exit 0 + unparseable stream", so both coordinates are pinned.
- **Human-mode witness**: same invalid flag without `--json` — exit 0 and the warning visible.
  Channel-agnostic by design: it survives the channel fix and fails only if the diagnostic is
  dropped entirely.
- **Hermeticity**: the fixture env now also blanks `LLMWIKI_COMPILE_CONCURRENCY` and points
  `LLMWIKI_CLAUDE_SETTINGS_PATH` at a nonexistent file, so ambient developer configuration
  cannot perturb the subprocess.

Mutation-tested both directions: restoring the stdout write turns the four channel witnesses
red (human-mode stays green, as designed); deleting the diagnostic turns all five red.
Full suite green (5033 passed), `tsc --noEmit` clean, fallow clean.

Based on `fix/quickstart-json-stdout` (#201) — merging order stays with #201.

### DIFF
--- a/test/quickstart-json-stdout.test.ts
+++ b/test/quickstart-json-stdout.test.ts
@@ -21,7 +21,7 @@ import { describe, expect, it } from "vitest";
 import { mkdtemp, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import path from "node:path";
-import { runCLI } from "./fixtures/run-cli.js";
+import { expectCLIExit, runCLI } from "./fixtures/run-cli.js";
 
 /** A workspace holding one local source, enough for quickstart to reach compile. */
 async function workspace(): Promise<string> {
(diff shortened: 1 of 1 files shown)


ethanj:
## Verdict: close #202 without merging. It is fully superseded by #204.

Reviewed at head `fe14864a` against `main` @ `cb4ff808`.

### The PR's own claims all check out

I verified each one rather than taking them on trust, and against the implementation that actually landed rather than the one this was written for:

- All five cases pass against #204's implementation, including `-1` (Commander does not swallow it as a flag).
- Mutation results reproduce exactly as stated: forcing stdout turns the four channel witnesses red and leaves human mode green; deleting the diagnostic turns all five red.
- The exit-code argument is correct. `runCLI` catches and returns rather than throwing, so without `expectCLIExit` a nonzero-exit regression with an intact envelope would have passed silently.
- `LLMWIKI_CLAUDE_SETTINGS_PATH` is a real override, read by `src/utils/claude-settings-reader.ts`, so the hermeticity addition is not cargo cult.

It also catches a failure mode it was not written for: mutating the `cli.ts` call site to forget `options.json ? "stderr" : "stdout"` turns the four json witnesses red. #204's conditional design puts the defect in two places instead of one, and this oracle covers both.

### Why it should still be closed

`main` already contains every bullet in the description. #204 shipped its own `test/quickstart-json-stdout.test.ts` with the same four-value `it.each`, the same `expectCLIExit(result, 0)`, the same two hermeticity env blanks, and a human-mode witness.

On four counts the landed version is stricter:

| | #202 | #204 on main |
| --- | --- | --- |
| Envelope shape | `.version === 1` | full key allowlist via `Object.keys(...).toEqual([...])` |
| Warning absent from stdout | not asserted | `expect(result.stdout).not.toContain("--concurrency")` |
| Warning present on stderr | `toContain` | exact equality on the stripped line |
| Human mode | channel-agnostic (`stdout + stderr`) | pins stdout **and** asserts stderr stays clean |

The human-mode difference is the one that matters. #202 argues channel-agnostic is deliberate so the witness survives the channel fix. Under #204 that reasoning no longer applies: human mode deliberately keeps stdout, so the landed assertion pins a real contract while the channel-agnostic version would pass even if human-mode output drifted to stderr.

The explicit `60_000` timeout also buys nothing. `vitest.config.ts` sets `testTimeout: 30_000` globally and each case runs in ~360ms.

### Structural problem

The base branch is `fix/quickstart-json-stdout` (#201), which was closed unmerged on 1 September. #202 therefore still carries `e2966a3e`, #201's unconditional-stderr implementation. Retargeting to `main` conflicts in two files:

```
CONFLICT (content): src/compiler/concurrency.ts
CONFLICT (add/add): test/quickstart-json-stdout.test.ts
```

The add/add is the tell: both efforts created a file at the same path. Resolving it means discarding this side, because the other side is the one that shipped.

### The one piece worth keeping

#202's file docblock explains the *mechanism*: the flag is parsed while the CLI assembles the command's arguments, so quiet mode inside the command body can never be early enough. Main's docblock states the contract but not why it cannot be fixed by moving `setQuiet` earlier. That paragraph is worth porting into the file on `main` as a comment, so the next person does not re-derive it. That is a two-line docs change, not a reason to keep this branch.

### Suggested action

Close #202, and delete both `test/quickstart-concurrency-coverage` and `fix/quickstart-json-stdout`, which are still on the remote.

PR #206, opened 2026-09-08
feat(embeddings): allow disabling refreshes

LorenzoGentile:
## Summary

Deployments that maintain their own vector index can disable llmwiki's embedding writes with `LLMWIKI_EMBEDDINGS=off`, `false`, `0`, or `no` (case-insensitive). Compile still validates its chat provider; query/search retain their embedding-provider checks. Disabled write paths leave embedding, pending, and quarantine files untouched.

Enabled non-review compiles reconcile missing or stale vectors even when no source changed, without regenerating source-derived pages or rewriting a healthy store. This applies to all enabled compiles, including projects that have never used the switch.

## Bounded reconciliation retries

- Automatically discovered work is recorded before provider calls and shares the existing five-failure budget with explicit page changes.
- Exhausted page IDs move from the active pending queue to `.llmwiki/quarantined-embeddings.json`, using the same bounded, root-confined file handling. Later unchanged refreshes exclude them, so they cannot retry indefinitely or block healthy pages.
- An explicit page change releases its quarantine and starts a fresh budget. Disabling refreshes or changing only embedding configuration does not reset quarantine. The environment reference explains recovery without regenerating pages.
- Compile and environment docs explain unchanged-compile scans, potential full re-embedding and provider costs after backend/model/endpoint changes, and strict-mode failure behavior.

## Relationship to released SDK support

The SDK-only `compile({ embeddings: false })` option shipped in v1.2.0. This PR adds the env/CLI contract, applies the opt-out to shared refresh/finalization paths, and reconciles missing or stale vectors when refreshes are enabled. Updated with upstream `main` at v1.3.0.

## Verification

- Node 24: `npx tsc --noEmit`, test type-check ratchet (no new diagnostics), build, and Fallow codebase analysis pass.
- Full suite: `npm test -- --maxWorkers=4` — 5,532 passed, 3 skipped.
- Real-core regressions cover quarantine across unchanged refreshes, bounded automatic backfill retries, write-ahead recording, healthy-page progress, explicit re-queue, interrupted settlement, strict-mode failure, and disabled-state preservation.
- Mutation check: bypassing retry preparation makes the quarantine regression fail with three provider calls instead of one; the restored implementation passes.
- Real compile regressions cover off-to-on missing/stale recovery, healthy-store no-op behavior, model changes without page regeneration, and strict-mode failures on unchanged sources.
- Built-CLI regression retains compile opt-out behavior while query still requires an embedding backend.
- Release-docs check and local preview of both updated documentation pages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,12 @@ Credit includes original proposals, reports and diagnoses, including work comple
 
 ### Added
 
+- **Embedding refresh opt-out** — `LLMWIKI_EMBEDDINGS=off` skips embedding
+  refreshes without reading or writing the durable pending queue. This lets
+  deployments with their own semantic index keep compile and review workflows
+  while avoiding an embedding store they never consume. Unset and other values
+  preserve the existing behavior.
+
 - **OpenAI Codex CLI provider** — `LLMWIKI_PROVIDER=codex-agent` or
   `--provider codex-agent` delegates chat and structured generation to a locally
   authenticated Codex CLI. The same one-run `--provider` override is now
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,8 +40,9 @@ Credit includes original proposals, reports and diagnoses, including work comple
 
 ### Added
 
-- **Embedding refresh opt-out** — `LLMWIKI_EMBEDDINGS=off` skips embedding
-  refreshes without reading or writing the durable pending queue. This lets
+- **Embedding refresh opt-out** — `LLMWIKI_EMBEDDINGS=off` skips all embedding
+  refreshes, including `query --save`, without reading or writing the embedding
+  store or durable pending queue. This lets
   deployments with their own semantic index keep compile and review workflows
   while avoiding an embedding store they never consume. Unset and other values
   preserve the existing behavior.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Embedding refresh opt-out** — `LLMWIKI_EMBEDDINGS=off`, `false`, `0`, or
+  `no` skips all embedding refreshes, including `query --save`, without reading
+  or writing the embedding store or durable pending queue. Compile still
+  validates its chat provider but does not require an embedding backend. This
+  lets deployments with their own semantic index keep compile and review
+  workflows while avoiding an embedding store they never consume.
+
+  Existing embedding and pending files remain untouched while refreshes are
+  disabled. After re-enabling them, the next non-review compile reconciles
+  missing and stale vectors even when no sources changed.
+
 ## [1.2.0] - 2026-09-09
 
 ### Highlights
@@ -39,13 +52,6 @@ Credit includes original proposals, reports and diagnoses, including work comple
 - Windows path fixes are included, but release CI remains Linux-only. Full Windows filesystem-confinement guarantees are not verified; use Linux for untrusted projects until #175 is resolved.
 
 ### Added
-
-- **Embedding refresh opt-out** — `LLMWIKI_EMBEDDINGS=off` skips all embedding
-  refreshes, including `query --save`, without reading or writing the embedding
-  store or durable pending queue. This lets
-  deployments with their own semantic index keep compile and review workflows
-  while avoiding an embedding store they never consume. Unset and other values
-  preserve the existing behavior.
 
 - **OpenAI Codex CLI provider** — `LLMWIKI_PROVIDER=codex-agent` or
   `--provider codex-agent` delegates chat and structured generation to a locally
--- a/docs/configuration/environment-variables.mdx
+++ b/docs/configuration/environment-variables.mdx
@@ -16,6 +16,7 @@ You don't need to set everything - only the variables relevant to your chosen pr
 |---|---|---|
 | `LLMWIKI_PROVIDER` | `anthropic` | Provider to use. One of: `anthropic`, `openai`, `ollama`, `minimax`, `copilot`, `claude-agent`, `codex-agent`, `orcarouter`, `atlascloud` (aliases: `atlas-cloud`, `atlas`) |
 | `LLMWIKI_MODEL` | Provider default | Model name to use. Overrides the provider's built-in default model. For `codex-agent`, unset delegates model choice to Codex and a set value is passed through with `--model` |
+| `LLMWIKI_EMBEDDINGS` | Enabled | Set to `off` (case-insensitive) to skip embedding refreshes during compile, review approval, and import. Existing embedding and pending files are left untouched. |
 | `LLMWIKI_EMBEDDING_PROVIDER` | Active chat provider | Backend serving embeddings, independent of `LLMWIKI_PROVIDER`. One of `anthropic`, `claude-agent`, `openai`, `ollama` |
 | `LLMWIKI_EMBEDDING_MODEL` | Provider default | Embedding model to use. Applies only when the effective embedding provider is `openai` or `ollama` - ignored for `anthropic` and `claude-agent`, even when `LLMWIKI_EMBEDDING_PROVIDER` names one of them explicitly |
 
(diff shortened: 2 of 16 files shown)


ethanj (changes requested):
Thanks for adding this. A narrow refresh opt-out is useful for projects that maintain their own vector index, and putting the checks in the shared refresh paths keeps the change small. The tests cover case-insensitive input, preservation of existing files, and the enabled path well.

Two behaviors need addressing before merge:

### 1. Re-enabling does not backfill pages compiled while disabled

The early return in [embeddings-refresh.ts](https://github.com/atomicstrata/llm-wiki-compiler/blob/1be77fdd2d50bf03f218ffdcdc50f2a163ffcf43/src/utils/embeddings-refresh.ts#L69-L80) skips recording the changed page IDs. Source state still advances, so an ordinary compile after removing the variable has neither changed pages nor pending IDs and returns before reaching the embedding writer.

Reproduced with the real compile pipeline: compile one source with `LLMWIKI_EMBEDDINGS=off`, unset the variable, then compile again without changing the source. The page exists, but the second compile makes zero embedding-writer calls. A missing index stays missing; changes made while disabled are not reconciled by that unchanged compile.

Could you define the re-enable behavior and add a regression covering this transition? Preferably an enabled refresh can discover missing or stale embeddings without regenerating source pages. If intentionally leaving them untouched is part of the narrow contract, please document that limitation and a concrete recovery procedure in the environment-variable reference. Recording pending work while disabled is another option, but would change the current no-writes promise and is not required here.

### 2. Compile preflight still requires an embedding backend when refreshes are off

The new checks run after the CLI's [provider guard](https://github.com/atomicstrata/llm-wiki-compiler/blob/1be77fdd2d50bf03f218ffdcdc50f2a163ffcf43/src/utils/provider-guard.ts#L101-L112), which does not account for the opt-out. With the CLI-backed chat provider and no explicit embedding backend configured, setting `LLMWIKI_EMBEDDINGS=off` still makes `compile` exit demanding an embedding backend, before the refresh guards can take effect. This reproduces through the built CLI.

Please let compile preflight honor an explicit refresh opt-out, with a CLI-level regression. Keep embedding validation for commands that actually need semantic retrieval; this need not become a blanket bypass of provider checks.

### Non-blocking usability notes

- The Sources-section switch now on main accepts `0`, `false`, `off`, and `no`; this switch accepts only `off`. Matching those spellings would reduce surprises. The current behavior is explicitly documented, so this is a consistency suggestion rather than another blocker.
- An existing pending marker continues to produce warnings in lint/status while refreshes are disabled. A short documentation note would clarify that leaving the file untouched also leaves its warning visible. There is no need to delete retry state as part of this change.

Verification: the four new tests and two existing pending-refresh tests passed. An independent off-to-unset compile regression fails on the missing writer call, and the CLI preflight failure reproduces separately. Type-check and build passed. Hosted checks are green at `1be77fdd`; the branch needs updating onto current main before the final gate.

Thanks again for the contribution. These asks can stay focused on the toggle's lifecycle and entry points without expanding the feature into an embedding-system redesign.

LorenzoGentile:
Addressed both blockers and the two usability notes in `9437396`, after rebasing the branch onto current `main` (`v1.2.0`).

1. Re-enable now reconciles through the existing embedding core even when the compile delta is empty. That core discovers missing/stale vectors by content hash; a healthy store remains a no-op with no provider call or write. Real compile-pipeline tests cover both a missing store and a page changed while refreshes were disabled, followed by an unchanged enabled compile.
2. Compile-specific provider preflight now omits embedding-only requirements when `LLMWIKI_EMBEDDINGS` or the SDK `embeddings: false` option disables refreshes. Chat-provider validation remains mandatory. Query/search still use the full guard. A built-CLI regression covers `codex-agent` compile without an embedding backend, plus the corresponding query failure.

The opt-out accepts `0`, `false`, `no`, and `off` case-insensitively. The environment-variable docs now explain the write-only scope, retained lint/status warning for an existing pending marker, and the recovery procedure after re-enabling.

Verification on Node 24: 5,347 tests passed (3 skipped), typecheck passed, build passed, and Fallow reported no dead code, duplication, or complexity-threshold violations. The repository pre-push hook repeated build and the full suite successfully before updating the branch.

LorenzoGentile:
Re-checked this after the v1.2.0 release. The release includes the SDK-only `compile({ embeddings: false })` option from #198; this PR still adds the env/CLI contract, applies it consistently to shared refresh/finalization paths, and reconciles missing or stale vectors when embeddings are re-enabled. I updated the PR description to make that boundary explicit. The branch is already rebased onto v1.2.0 and the hosted checks are green.

ethanj (changes requested):
Thanks for addressing the earlier feedback. Both original blockers are fixed: re-enabling refreshes reconciles missing or stale vectors without regenerating pages, and compile preflight respects the opt-out while query/search retain their embedding checks. The accepted spellings and pending-marker documentation are covered too.

Two remaining items before merge:

### 1. Preserve the retry limit during automatic reconciliation

In [embeddings-refresh.ts:82–99](https://github.com/atomicstrata/llm-wiki-compiler/blob/943739681230dab30865fb046b431f369feb342b/src/utils/embeddings-refresh.ts#L82-L99), the core now runs even when the pending queue is empty. That also rediscovers pages whose embedding attempts were just quarantined.

Reproduced with the real refresh core:

- Start an eligible page at four failed attempts.
- The next failure reaches the five-attempt limit, removes its pending entry, and reports that it “will not be retried.”
- Two subsequent unchanged refreshes nevertheless call the failing provider again.

These later attempts have no pending entry, so they bypass the retry limit indefinitely. Restoring the old empty-queue guard makes this regression test pass, but would reopen the re-enable problem.

Please retain automatic backfill while preserving bounded retries, with a regression covering unchanged refreshes after quarantine.

### 2. Explain automatic reconciliation in the user-facing docs

The current wording presents reconciliation as recovery after using the switch, but it also changes behavior for users who never disable embeddings:

- Empty-delta refreshes now scan eligible pages and the embedding store.
- Changing the embedding model or provider can trigger a full re-embed on an unchanged compile, with associated provider costs.
- With `LLMWIKI_EMBED_STRICT`, that otherwise unchanged compile can now fail if embedding generation fails.

This can be useful behavior; users just need to know about it where they configure and run the feature. Please explain unchanged-compile reconciliation and the potential full re-embed in the compile docs, and cover embedding-configuration changes and strict-mode failure behavior in the environment-variable reference. A short CHANGELOG entry can announce the change and link to that guidance, rather than being its only home. No new documentation page is needed, and optimizing the scan can remain separate work.

The original fixes have passing regression and mutation evidence. At this head, 55 focused tests, type-check, and build passed, and hosted CI is green. The independent quarantine regression exposes the remaining retry gap.

Thanks again—the follow-up closes the original issues. Preserving the retry boundary and documenting the wider behavior would complete this change.

LorenzoGentile:
Addressed both remaining items in `587bacb`, and merged current upstream `main` (v1.3.0) to resolve the conflicts.

1. **Bounded retries during automatic reconciliation.** Automatically discovered missing/stale vectors now enter the durable pending queue before provider calls and share the existing five-failure budget. Exhausted IDs move to `.llmwiki/quarantined-embeddings.json`, using the same bounded, root-confined file handling. Subsequent unchanged refreshes exclude them while allowing healthy pages to embed. An explicit page change releases that page's quarantine with a fresh budget; disabling refreshes or changing embedding configuration alone does not reset it.

   Real-core regressions cover unchanged refreshes after quarantine, backfill failures with no initial pending entry, write-ahead recording, healthy-page progress, explicit re-queue, interrupted settlement, strict-mode failures, and preservation while disabled. As a mutation check, bypassing retry preparation makes the quarantine regression fail with three provider calls instead of one; the restored implementation passes.

2. **User-facing reconciliation guidance.** The compile guide now explains that enabled non-review compiles scan pages and the store even with unchanged sources, including for users who have never disabled embeddings. It covers potential full re-embedding and provider costs after configuration changes. The environment reference documents provider/model/endpoint changes, strict-mode failures on otherwise unchanged compiles, quarantine, and recovery without regenerating pages. The changelog links to this guidance. Real compile tests also cover a model change without page regeneration and a strict-mode failure during unchanged-source backfill.

Verification on Node 24: **5,532 tests passed, 3 skipped**; source type-check, test type-check ratchet, build, Fallow, release-docs check, and local preview of both updated guides passed. All three hosted checks are green at `587bacb`: `build-and-test (24)`, `codebase-health`, and `test-typecheck`.

Thanks for the follow-up review—ready for another look.

ethanj (approved):
Thanks for the contribution, Lorenzo, and for your patience working through the review feedback. I appreciate the care you put into the implementation, regression tests, and documentation. Glad to have this in llmwiki!

PR #229, opened 2026-09-19
perf(viewer): resolve wikilinks through a slug index instead of rescanning pages per link

chenleji:
## Problem

`resolveBareSlug` resolves one wikilink target by scanning the whole page list up to four times — `concepts` slug, `queries` slug, `concepts` alias, `queries` alias — and every scan `slugify()`s each page's aliases. It runs **once per wikilink occurrence**, so the cost is O(links × pages × aliases).

On a 2356-page wiki with ~13.8k wikilinks that is tens of millions of `slugify` calls, and it dominates `llmwiki context`:

```
node --cpu-prof … llmwiki context --json "<question>"
  30.1% + 29.7%  slugify                        (@ cli.js bundle, both threads)
   5.4% +  5.3%  RegExp: [^\p{L}\p{N}\s-]       (inside slugify)
   3.7% +  3.5%  resolveBareSlug
   3.2% +  3.1%  (anonymous, the alias matcher inside resolveBareSlug)
   1.0% +  0.9%  (anonymous, the exact-slug matcher)
```

`llmwiki context` took **1m50s** on that wiki, which is long enough that consumers treat retrieval as unusable and fall back to `grep`.

## Change

Resolve through four lookup tables built **once per page list** (`WeakMap` keyed on the list's identity, so the index is shared by every lookup in one parse and goes away with the list):

- `conceptBySlug`, `queryBySlug`, `conceptByAlias`, `queryByAlias`
- first page in list order wins per bucket, which preserves the previous `.find()` semantics and the existing precedence (exact concept → exact query → concept alias → query alias)
- `hasAliasSlug` becomes unused and is removed

Also in this commit: the freshness snapshot's `source → concept` lookup inverted the same way. `ownersOf` did `Object.values(snapshot.sources).filter(s => s.concepts.includes(slug))` **per page** (2356 pages × 289 sources per snapshot); it now reads a reverse index built once per snapshot.

## Verification

`llmwiki context --json`, same 2356-page wiki, same question:

| | before | after |
|---|---|---|
| wall time | 1m50s | **4.7s** |
| `primary[]` entries | 5 | 5 (identical ids, same order) |

- `llmwiki lint` on the full wiki: output is **byte-identical** to the pre-change build (10508 lines; 10308 error / 202 warning / 13 info). This is the strongest available check that link resolution, alias precedence and dangling-link classification are unchanged.
- `npx tsc --noEmit` → 0 errors; `npm run build` → success.
- `vitest run` on the link/viewer surface: 4 files, 91 tests pass.
- `fallow` (the repo's pre-commit health analyzer, v2.82.0): **0 issues above threshold**, `16564 analyzed · maintainability 90.3 (good)`, exit code 0 — the changes add no dead code, duplication or complexity.
- Freshness is unaffected in every way the CLI reports it: `llmwiki status` prints the same aggregates before and after (`2356 concepts`, `289 sources`, `no stale or orphaned pages`, `State: ok`), and the byte-identical `lint` output above includes every `stale-page` warning the freshness layer would emit. (The reverse index is a pure lookup change; it returns the same owner set for a slug, just without rescanning.)

## Notes

- No behavior change, no output change, no storage-format change — this is purely an algorithmic fix.
- Independent of any retrieval-semantics work; it can be reviewed on its own.
- The remaining cost on a large wiki is dominated by page parsing, so this is close to the floor for the current design.


ethanj (changes requested):
Thanks for this! Indexing the repeated lookups is a useful improvement. The freshness reverse index looks sound, but there is one behavior regression to address before merging.

### Keep typed records out of the concept index

In [collect.ts:145–149](https://github.com/atomicstrata/llm-wiki-compiler/blob/6e72027f7aab0785b4f68c0a54fbe167bf3cedcc/src/viewer/collect.ts#L145-L149), every page whose directory is not `queries` is inserted into the concept buckets. The previous resolver explicitly matched only `concepts` and `queries`.

This is reachable in profile-enabled wikis because the renderer passes the whole snapshot, including typed records. With `papers/alpha` present, `[[alpha]]` becomes a live link to nonexistent `concepts/alpha`. If `queries/alpha` also exists, the typed paper incorrectly shadows that valid query. Both focused reproductions fail with the new index and pass with main's previous resolver; default-profile tests do not expose this difference.

Could you skip entries whose directory is neither `concepts` nor `queries`, and add regression coverage for a typed-only slug and a typed record sharing a query's slug? The optimization and existing precedence can otherwise stay intact.

Non-blocking: the exported `resolveBareSlug` now caches by array identity, so changing the array or its entries after a lookup leaves cached keys stale. No in-tree caller was found doing that. Please document that the page list and resolution fields must remain unchanged for the cache lifetime, or explicitly define how callers invalidate it; this does not need a larger cache redesign.

Reviewed at `6e72027f`. The merge update did not change the relevant implementation; the earlier focused evidence remains applicable. Thanks again for the performance work.

PR #230, opened 2026-09-19
fix(mcp): decode percent-encoded slugs when reading resources (non-ASCII pages were unreadable)

chenleji:
## Problem

`resources/read` cannot return any page whose slug is not pure ASCII. On a wiki with non-ASCII page names (e.g. any Chinese knowledge base) this is effectively every page:

```
llmwiki://concept/agent-api-gateway          → OK
llmwiki://concept/ai理赔审核覆盖率与准确率提升目标
  → -32603 Page not found: wiki/concepts/ai%E7%90%86%E8%B5%94%E5%AE%A1%E6%A0%B8%E8%A6%86%E7%9B%96%E7%8E%87...
```

The lookup used the percent-encoded string as the filename. Sending an already-encoded URI fails identically, so clients cannot work around it.

Cause: resource URIs are built with `encodeURIComponent(slug)` in `listPagesUnder` (that part is deliberate — see the S13 comment about `#` and spaces), and the MCP SDK matches `ResourceTemplate` variables against the **normalised request path**, so the `{slug}` variable arrives still encoded. The read callbacks passed `String(slug)` straight into `loadPageWithMeta`, and the comment above the encoder assumed the read side decoded it back — it does not.

Impact beyond a single client: MCP resources are the only page-read path for hosts that do not surface tool calls (the Codex CLI, for instance, exposes `resources/read` and not the tool list), so those agents saw "server exposes resources only" and fell back to reading markdown off disk.

## Change

- add `decodeSlug`, which percent-decodes a template variable and returns the raw value when the escaping is invalid (a slug containing a literal `%`, e.g. `50% off`).
- use it in both templated read callbacks (`wiki-concept`, `wiki-query`).
- correct the encoder's comment, which documented the wrong assumption.

## Verification

`resources/read` against a 2356-page wiki, before → after:

| request | before | after |
|---|---|---|
| `llmwiki://concept/agent-api-gateway` (ASCII) | OK | OK |
| `llmwiki://concept/<中文 slug>` | `Page not found: …ai%E7%90%86…` | **page returned** (`slug`, frontmatter, body) |
| `llmwiki://concept/<percent-encoded 中文 slug>` | `Page not found: …` | **page returned** |

Tests — `test/mcp-server.test.ts`, 27 passing:

- the existing S13 test asserted the *correct* behaviour but passed `{ slug }` already decoded, so it never exercised the SDK's encoding; it now passes the encoded variable (this is the change that would have caught the bug).
- new: non-ASCII (CJK) slug resolves through the encoded URI.
- new: a slug with a literal `%` (`50% off`) still resolves, proving the decode falls back rather than throwing.

`npx tsc --noEmit` → 0 errors · `npm run build` → success · `fallow` → no issues, 0 duplicated lines, dead exports 0.0%.

## Notes

- No API, storage or output-format change; ASCII slugs behave exactly as before (the decode is the identity for them).
- Pre-existing behaviour kept: a malformed escape is not an error, it just means "the raw value is the slug".


ethanj (changes requested):
Thanks for fixing the Unicode resource reads. Decoding the template variable is needed, but it also opens a path boundary that needs protecting before this lands.

### Confine the decoded slug before reading a file

In [resources.ts:177–192](https://github.com/atomicstrata/llm-wiki-compiler/blob/f876c37fdf7c677448d8a087f4da1fe5cfc8fefc/src/mcp/resources.ts#L177-L192), decoding turns `%2F` into a path separator, then `loadPageWithMeta` joins it into a filename without confinement.

A reproduction through the actual MCP client/server and template matching returned a sibling test `secret.md` outside the project for `llmwiki://concept/..%2F..%2F..%2Fsecret`. The same request is rejected before this change. Both concept and query resources share the reader.

Please validate the decoded slug as a single safe filename component, rejecting separators and traversal, and use the existing confined wiki-reader path to enforce containment, including symlinks. Keep the working Unicode, spaces, hash-sign and literal-percent cases. Add actual-protocol tests proving encoded traversal cannot read outside the permitted wiki directory.

### Make the invalid-escaping test exercise its claimed branch

The new `50% off` test goes through `encodeURIComponent`, so the callback receives `50%25%20off`, which decodes successfully. It verifies a valid literal-percent round trip, not the catch/fallback branch. Please label that case accordingly and add a genuinely malformed-escape case if raw fallback is part of the intended contract. Driving it through SDK URI matching will also pin the actual page-not-found behavior rather than assuming which variable the SDK supplies.

Reviewed at `f876c37f`. The merge update did not change this implementation. The 27 existing MCP tests passed in the earlier focused run, but the added confinement witness fails. Thanks again for addressing the resource-read gap.

PR #231, opened 2026-09-21
feat(review): approve candidate batches with one finalization

LorenzoGentile:
## Summary

Add `llmwiki review approve-batch --input <manifest.json> --json` so callers can approve a bounded set of reviewed candidates while paying the global wiki finalization cost once instead of once per candidate.

The command:

- accepts a versioned manifest with candidate IDs and optional draft-content hashes;
- acquires the project lock once, recovers pending writes, and validates every candidate under that lock;
- applies the valid subset, then runs source-state persistence, link resolution/repair, index/MOC generation, and scoped embedding refresh once;
- returns stable per-candidate `approved`, `invalid`, `conflict`, or `failed` outcomes;
- retains candidates when shared finalization fails, allowing an idempotent retry;
- keeps the existing single-candidate approval command, now using the same shared finalizer.

Inputs are capped at 100 candidates and 1 MiB. The planner rejects duplicate hash pins, stale content, source-snapshot conflicts, relation-gated typed promotions, and portable case/Unicode target aliases before writes. The portable comparison intentionally treats names such as `Straße` and `STRASSE` as aliases because the default macOS filesystem resolves them to the same file.

## Why

Single approval currently repeats whole-wiki work for every candidate. In the downstream integration that motivated this change, approving 30 candidates in a 1,500-page wiki took 64.35 seconds one by one versus 5.36 seconds through this batch path (about 12x faster). The included benchmark script creates the same disposable corpus for both modes.

## Recovery semantics

Known candidate validation problems produce individual results and do not prevent unrelated candidates from proceeding. Unexpected storage errors or shared-finalization failures fail the batch, leave candidates available for retry, and preserve embedding reconciliation intent. Page mutations continue to use the existing write journal and recovery path.

## Validation

- `npx tsc --noEmit`
- `npm run typecheck:tests` — 1,347 files checked; no baseline regression
- `npm run build`
- `npm run release:check-docs`
- `npm run fallow:ci` — no dead code, duplication, or threshold violations
- `vitest run test/review-batch-targets.test.ts` — 21/21 passed
- full suite — 5,677 passed, 3 skipped; three unrelated `workflow-active-run-quota` cases exceeded their fixed 30-second timeout on the local macOS filesystem
- isolated quota suite with a 60-second timeout — 5/5 passed; the three slow cases completed in 22–27 seconds each without suite contention

CI on Node 24/Linux remains the canonical full-suite check.


ethanj:
Thanks, Lorenzo — batching the shared finalization is a useful improvement, and I appreciate the recovery tests you've included.

Sorry for the moving target here. We were in the middle of merging the large #233 change when you submitted this, and it has now landed on public main. That introduced some new approval and recovery safeguards, so this PR needs a rebase plus two behavioral adjustments. These are integration gaps with the newly merged code, not requirements you missed in the base you worked from.

I reviewed head `917b12a7`; all 87 tests in the eight focused batch/recovery/embedding suites I ran passed. I also exercised your batch implementation against project state produced by current main. Those compatibility checks exposed the following:

### 1. Preserve the new target-page preconditions during batch approval

In [review-batch-plan.ts](https://github.com/atomicstrata/llm-wiki-compiler/blob/917b12a7bf1151f9dc7978d33bd1539f4b029f1b/src/commands/review-batch-plan.ts#L75-L82), the candidate body pin is checked, but `expectedTargetHash` and `expectTargetAbsent` are not. Current main's single approval checks these through `targetUnchangedSincePropose` before writing.

Reproduction: stage a repair with current main's `lint --fix-propose`, edit the live page afterward, then approve. Single approval correctly refuses and preserves the edit; the batch implementation reports `approved` and overwrites the edit with the stale repair. A candidate-body hash alone cannot detect that the destination changed.

Could you share the same under-lock target-precondition check between single and batch approval, returning an individual `invalid` result for a mismatch? Regression cases should cover both a changed target hash and a page appearing after a proposal that required the target to remain absent.

### 2. Enter through the new mutation recovery gate

In [review-approve-batch.ts](https://github.com/atomicstrata/llm-wiki-compiler/blob/917b12a7bf1151f9dc7978d33bd1539f4b029f1b/src/commands/review-approve-batch.ts#L53-L58), raw `acquireLock` followed by page-journal recovery does not run the operation-bundle and preparation recovery checks now used by the other mutation commands.

With a parked operation bundle, current main's `acquireMutationLock(root, "review")` refuses with `bundle-recovery-blocking`; the batch implementation on the same project proceeds and returns `completed`. That permits unrelated writes while recovery still needs the project's existing state.

Please route batch entry through the shared review mutation-lock gate and retain the current-main candidate-store custody checks before live writes. A regression should verify that a recovery-blocked project remains unchanged and its candidates stay pending.

### Rebase details

The current conflicts are in `package.json`, `src/cli/review-commands.ts`, `src/trust/promote.ts`, `src/utils/embeddings-refresh.ts`, and `src/utils/embeddings.ts`. When reconciling the embedding changes, please also preserve main's `pruned`-ID settlement behavior so successfully removed vectors do not remain queued for retries.

The compatibility checks above are not a validation of a resolved merged tree; the rebased version will need another pass. Thanks again for the contribution, and apologies for the extra integration work caused by the timing of our merge. Happy to help with any of the overlap.

ethanj (changes requested):
Thanks, Lorenzo — batching the shared finalization is a useful improvement, and I appreciate the recovery tests you've included.

Sorry for the moving target here. We were in the middle of merging the large #233 change when you submitted this, and it has now landed on public main. That introduced some new approval and recovery safeguards, so this PR needs a rebase plus the two integration adjustments below. Those are gaps with the newly merged code, not requirements you missed in the base you worked from. An additional review also found two batch-recovery cases on the submitted head, included as items 3 and 4.

I reviewed head `917b12a7`; all 87 tests in the eight focused batch/recovery/embedding suites I ran passed. I also exercised your batch implementation against project state produced by current main. Those compatibility checks exposed the following:

### 1. Preserve the new target-page preconditions during batch approval

In [review-batch-plan.ts](https://github.com/atomicstrata/llm-wiki-compiler/blob/917b12a7bf1151f9dc7978d33bd1539f4b029f1b/src/commands/review-batch-plan.ts#L75-L82), the candidate body pin is checked, but `expectedTargetHash` and `expectTargetAbsent` are not. Current main's single approval checks these through `targetUnchangedSincePropose` before writing.

Reproduction: stage a repair with current main's `lint --fix-propose`, edit the live page afterward, then approve. Single approval correctly refuses and preserves the edit; the batch implementation reports `approved` and overwrites the edit with the stale repair. A candidate-body hash alone cannot detect that the destination changed.

Could you share the same under-lock target-precondition check between single and batch approval, returning an individual `invalid` result for a mismatch? Regression cases should cover both a changed target hash and a page appearing after a proposal that required the target to remain absent.

### 2. Enter through the new mutation recovery gate

In [review-approve-batch.ts](https://github.com/atomicstrata/llm-wiki-compiler/blob/917b12a7bf1151f9dc7978d33bd1539f4b029f1b/src/commands/review-approve-batch.ts#L53-L58), raw `acquireLock` followed by page-journal recovery does not run the operation-bundle and preparation recovery checks now used by the other mutation commands.

With a parked operation bundle, current main's `acquireMutationLock(root, "review")` refuses with `bundle-recovery-blocking`; the batch implementation on the same project proceeds and returns `completed`. That permits unrelated writes while recovery still needs the project's existing state.

Please route batch entry through the shared review mutation-lock gate and retain the current-main candidate-store custody checks before live writes. A regression should verify that a recovery-blocked project remains unchanged and its candidates stay pending.

### 3. Check known recovery-storage prerequisites before promoting pages

In [review-finalize.ts:53–56](https://github.com/atomicstrata/llm-wiki-compiler/blob/917b12a7bf1151f9dc7978d33bd1539f4b029f1b/src/commands/review-finalize.ts#L53-L56), the intent file is first opened and recorded after the whole approval page batch has already committed.

With malformed `.llmwiki/review-embedding-intent.json`, a focused reproduction returns `failed`, but the approved page is already live and source-state finalization has not even started. Retrying repeats promotion before encountering the same known storage problem.

Could you open/validate the intent store and persist the initially known candidate-page intent under the lock before promotion? This does not require making the entire finalization tail transactional; it just avoids live writes when a prerequisite is already known to be unusable. Keep recording additional collateral IDs before their link writes, and add a corrupt-intent regression that asserts no candidate page was promoted. Capacity checks for already-known work can similarly happen early, without assuming all collateral work is known in advance.

### 4. Do not spend provider retry budgets on a required store migration

In [embeddings-refresh.ts:103–112](https://github.com/atomicstrata/llm-wiki-compiler/blob/917b12a7bf1151f9dc7978d33bd1539f4b029f1b/src/utils/embeddings-refresh.ts#L103-L112), a legacy or incompatible store refused by the affected-only planner enters the generic failure handler and increments attempts, even though no provider request was made. The refresh still reports a successful durable handoff when there is no marker overflow, allowing the batch to finish.

A focused reproduction with a legacy store performed five affected-only refreshes of the same page: zero provider calls, but the page ended up quarantined with five attempts. Normal reconciliation excludes quarantined IDs, so a migration prerequisite can turn into an exclusion from the very recovery it needs.

Please distinguish “full reconciliation required” from an attempted embedding failure: retain recoverable intent and an actionable diagnostic without consuming or exhausting the page's provider-attempt budget. Add a regression proving repeated migration-blocked refreshes do not quarantine the page and that a subsequent full reconciliation can recover it.

### Rebase details

The current conflicts are in `package.json`, `src/cli/review-commands.ts`, `src/trust/promote.ts`, `src/utils/embeddings-refresh.ts`, and `src/utils/embeddings.ts`. When reconciling the embedding changes, please also preserve main's `pruned`-ID settlement behavior so successfully removed vectors do not remain queued for retries.

The compatibility checks above are not a validation of a resolved merged tree; the rebased version will need another pass. Thanks again for the contribution, and apologies for the extra integration work caused by the timing of our merge. Happy to help with any of the overlap.

